PR #6, opened 2022-01-10
Add JumpCloud Agent

jb3:
This PR adds a role to install the JumpCloud agent to all our hosts.

It will pull the bash script from the JumpCloud website using the secret stored in the vault.

It checks whether the `jcagent.service` service is located on the host before running the installation script.


ChrisLovering (approved):
Nice, first try, every time

PR #7, opened 2022-01-10
Correct GitHub workflow step names

ChrisLovering:


HassanAbouelela (approved):
These are just like… your opinions man

PR #10, opened 2022-01-11
WireGuard

jb3:
This PR introduces a `wireguard` role which:
- Installs wireguard, wireguard-tools and linux-headers to hosts
- Generates a public and private key (if necessary)
- Templates a configuration file with all other host public keys and endpoints
- Ensures the WireGuard service is started and enabled

![mesh in action][https://media.discordapp.net/attachments/675756741417369640/930259819791806484/unknown.png?width=659&height=731]


ChrisLovering (approved):
:joe_salute:

PR #11, opened 2022-01-11
Resolve FQDNs in /etc/hosts

jb3:
PR #8 swapped from hardcoded IPs in the Ansible Inventory to using FQDNs on `pydis.wtf`. This change worked fine and does the exact same thing from an Ansible side, but we were using the `ansible_host` variable in the /etc/hosts template file controlled by the common role.

This meant that when templated instead of receiving an IP where it expected an IP it received an FQDN, leading to an invalid /etc/hosts file being templated to the servers. This change fixes that by resolving the `pydis.wtf` FQDN (e.g. `turing.box.pydis.wtf`) to it's resulting A record, and then injects that to the configuration file (the FQDN is injected afterwards anyway as is expected with /etc/hosts)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 ansible
 ansible-lint[yamllint]
 pre-commit
+dnspython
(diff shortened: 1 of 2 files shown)


jb3:
More info on the `dig` lookup module [here](https://docs.ansible.com/ansible/latest/collections/community/general/dig_lookup.html).

PR #16, opened 2022-01-11
Add basic UFW rules

### LINKED ISSUE #9: UFW Setup
# Abstract

We should setup UFW on our nodes to control inbound and outbound traffic.

# Implementation

There are several ways to implement this so we should talk about whether roles should control their own UFW rules or whether we should have a ufw role which manages all rules.

Ansible has a [`ufw`](https://docs.ansible.com/ansible/latest/collections/community/general/ufw_module.html) module for rule management.

We need to allow SSH, Wireguard and HTTP off the top of my head, there are probably other things here and there as well.

MarkKoz:
Resolves #9

Rules included:

- Default deny policy
- SSH allowed
- WireGuard allowed

Other service-specific rules are loaded from `roles/ufw/vars/main.yml`, which will have a list of dictionaries, where each dict corresponds to options for the ufw module.


MarkKoz:
I'm not sure if this will work or how to test it

https://github.com/python-discord/infra/pull/16/files#diff-68eba24c77e511acfa744d0aeb3f24a1c7b57d56dcb27f455ba0500ceab3eeebR18-R20

jb3 (approved):

PR #17, opened 2022-01-11
Add postgres role

### LINKED ISSUE #13: Set up PostgreSQL primary
# Abstract

We need to migrate our PostgreSQL database to the new hosts. As per PDEP-0001 this should reside on `lovelace`.

# Implementation

We will not be hosting the database within a container, and so should follow the regular installation processes for PostgreSQL on Debian 11. Most of this should just be interaction with the package manager, though we can use the [PostgreSQL modules for Ansible](https://docs.ansible.com/ansible/latest/collections/community/postgresql/index.html) to do some bootstrapping of users and databases.

Down the line we will also establish a read-only replica of PostgreSQL on `hopper`, but that is out of scope for this issue.

ChrisLovering:
Closes #13

This installs postgres, creates users and databases. What isn't supported is creating users with set passwords, currently only users with local auth can be made.

To support this we will need to create a vault file with those user's passwords.

### DIFF
--- a/roles/postgres/handlers/main.yml
+++ b/roles/postgres/handlers/main.yml
@@ -1,0 +1,4 @@
+- name: Restart postgres.
+  service:
+    name: '{{ postgresql_daemon }}'
+    state: "restarted"
--- a/roles/postgres/tasks/main.yml
+++ b/roles/postgres/tasks/main.yml
@@ -1,0 +1,14 @@
+- name: Install postgres packages.
+  apt:
+    name:
+      - python3-psycopg2
+      - postgresql-{{ postgresql_version }}
+      - postgresql-contrib-{{ postgresql_version }}
+      - libpq-dev
+    state: present
+
+- name: Check postgres is started and enabled on boot.
+  service:
+    name: '{{ postgresql_daemon }}'
+    state: started
+    enabled: true
--- a/roles/postgres/tasks/main.yml
+++ b/roles/postgres/tasks/main.yml
@@ -12,3 +12,20 @@
     name: '{{ postgresql_daemon }}'
     state: started
     enabled: true
+
+- name: Add postgres users.
+  community.postgresql.postgresql_user:
+    name: "{{ item.name }}"
+    login_user: "{{ item.login_user }}"
+    role_attr_flags: "{{ item.role_attr_flags }}"
+  with_items: "{{ postgresql_users }}"
+  become: true
+  become_user: "{{ postgresql_user }}"
+
+- name: Add postgres databases.
+  community.postgresql.postgresql_db:
+    name: "{{ item.name }}"
+    owner: "{{ item.owner | default(postgresql_user) }}"
+  with_items: "{{ postgresql_databases }}"
+  become: true
+  become_user: "{{ postgresql_user }}"
--- a/roles/postgres/tasks/main.yml
+++ b/roles/postgres/tasks/main.yml
@@ -14,18 +14,13 @@
     enabled: true
 
 - name: Add postgres users.
-  community.postgresql.postgresql_user:
-    name: "{{ item.name }}"
-    login_user: "{{ item.login_user }}"
-    role_attr_flags: "{{ item.role_attr_flags }}"
+  community.postgresql.postgresql_user: "{{ item }}"
   with_items: "{{ postgresql_users }}"
   become: true
   become_user: "{{ postgresql_user }}"
 
 - name: Add postgres databases.
-  community.postgresql.postgresql_db:
-    name: "{{ item.name }}"
-    owner: "{{ item.owner | default(postgresql_user) }}"
+  community.postgresql.postgresql_db: "{{ item }}"
   with_items: "{{ postgresql_databases }}"
   become: true
   become_user: "{{ postgresql_user }}"
--- a/roles/postgres/vars/main.yml
+++ b/roles/postgres/vars/main.yml
@@ -1,0 +1,7 @@
+postgresql_version: "13"
+postgresql_daemon: "postgresql@{{ postgresql_version }}-main"
+postgres_user: "postgres"
+
+postgresql_users: []
+
+postgresql_databases: []
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -4,3 +4,6 @@ vault_password_file = ./vault_passwords
 
 [privilege_escalation]
 become = yes
+
+[connection]
+pipelining = True
--- a/playbook.yml
+++ b/playbook.yml
@@ -21,3 +21,9 @@
     - wireguard
   tags:
     - wireguard
+
+- hosts: lovelace
+  roles:
+    - postgres
+  tags:
+    - postgres


roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: Add postgres databases.
+  community.postgresql.postgresql_db:
+    name: "{{ item.name }}"
+    owner: "{{ item.owner | default(postgresql_user) }}"
+  with_items: "{{ postgresql_databases }}"
+  become: true
+  become_user: "{{ postgresql_user }}"
+  vars:
+    ansible_ssh_pipelining: true
```
MarkKoz, line 61: Is there a way to avoid writing this 5 times, or would that involve a bunch of boilerplate that would make it not worthwhile?

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: Check postgres is running.
+  command: "{{ postgresql_bin_path }}/pg_ctl -D {{ postgresql_data_dir }} status"
+  changed_when: false
+  become: true
+  become_user: "{{ postgresql_user }}"
```
MarkKoz, line 67: What's the use of this?

roles/postgres/vars/main.yml
```diff
@@ -0,0 +1,14 @@
+postgresql_version: "14"
```
MarkKoz, line 1: How is 14 being installed? If it's just using apt, and we're on Debian stable, then the version would be 13.

If it depends on the Debian version, is there a nicer way to get the pg version than hard-coding it. Otherwise, upgrading Debian would entail remembering to update this variable too.

MarkKoz:
Also, pg needs to be assigned an IP and then allowed in ufw, right? Or is that out of scope for this PR?

roles/postgres/tasks/main.yml
ChrisLovering: Yea, it can
> Pipelining can be enabled globally by setting “pipelining=True” in the ssh section of ansible.cfg
However it has a drawback
> Unfortunately, pipelining is disabled by default because it is incompatible with sudo's requiretty setting (or su, which always requires a tty). This is because of a quirk of the Python interpreter, which enters interactive mode automatically when you pipe in data from a (pseudo) tty.

roles/postgres/tasks/main.yml
ChrisLovering: Afaik sudo doesn't have a role on psql by default, so this runs it under the `postgres` user

roles/postgres/vars/main.yml
ChrisLovering: hmm yea, we could either track the most recent version in debian, or we could add the postgres apt repo, which works similar to deadsnakes's ppa

roles/postgres/tasks/main.yml
MarkKoz: I highlighted multiple lines not just the last one. I'm asking why this entire step is necessary.

roles/postgres/tasks/main.yml
ChrisLovering: Volcyy mentioned yesterday that "Modern Debian comes with requiretty disabled", so I think we can actually enable this globally to get the benefit everywhere.

At that point the only duplicated code is that of the first two lines. I'm not sure how much of an issue this is, nor how we would go about aliasing the become steps.

roles/postgres/tasks/main.yml
ChrisLovering: It's mostly useful on subsequent runs of the role, when all of the above is skipped due to already existing, this ensures psql is still running

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
+- name: Install postgres Python library.
+  apt:
+    name: "python3-psycopg2"
+    state: present
+
+- name: Install postgres packages.
+  apt:
+    name:
+      - postgresql
+      - postgresql-contrib
```
jchristgit, line 10: Do we need this package? If yes, we should probably instead install:

```suggestion
      - postgresql-contrib-{{ postgresql_version }}
```

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
+- name: Install postgres Python library.
+  apt:
+    name: "python3-psycopg2"
+    state: present
```
jchristgit, line 4: This can be integrated with the task below.

roles/postgres/handlers/main.yml
```diff
@@ -0,0 +1,5 @@
+- name: Restart postgres.
+  service:
+    name: postgresql
```
jchristgit, line 3: This should use `postgresql_daemon` to select the proper cluster.

```suggestion
    name: '{{ postgresql_daemon }}'
```

roles/postgres/vars/main.yml
```diff
@@ -0,0 +1,11 @@
...
+postgresql_daemon: "postgresql@{{ postgresql_version }}-main"
+postgres_user: "postgres"
+
+postgresql_users:
+  - name: joe
+    login_user: joe
+    role_attr_flags: SUPERUSER
```
jchristgit, line 8: Do we want to keep this, or set it to an empty list for now?

```suggestion
postgresql_users: []
```

roles/postgres/vars/main.yml
```diff
@@ -0,0 +1,11 @@
...
+
+postgresql_databases:
+  - name: pythondiscord
```
jchristgit, line 11: Do we want to keep this? What would be stored on the database?

```suggestion
postgresql_databases: []
```

jchristgit (changes requested):
# Chris, this is absolutely excellent. One shudders to imagine a greater introduction into our Ansible repository than building the role for the World's Most Advanced Open Source Database, PostgreSQL. I have a few small nitpicks. They should not be much to implement. Feel free to ask me or Joe if you need help.

roles/postgres/handlers/main.yml
```diff
@@ -0,0 +1,5 @@
...
+    state: "restarted"
+    sleep: 5
```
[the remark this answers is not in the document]
ChrisLovering: This possibly isn't needed now, as we don't have an init scripts that could cause odd behaviour by exiting immediately on getting a stop signal.

I'm not sure how likely we are to add those, so I think this could be safely removed.

jchristgit (approved):
Absolutely great, rock solid, infra structure code. Thank you Chris. I could not expect any better, and was still positively surprised. Thank you. Thank you.

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: Check postgres is started and enabled on boot.
+  service:
+    name: '{{ postgresql_daemon }}'
+    state: started
+    enabled: true
+
+- name: Add postgres users.
+  community.postgresql.postgresql_user:
+    name: "{{ item.name }}"
```
MarkKoz, line 18: Instead of manually mapping each value, you can expand the entire dict. See my ufw PR for reference.

MarkKoz (changes requested):

roles/postgres/tasks/main.yml
ChrisLovering: Oh, this is very nice, thanks! Should help us if we want to add more in future, since the built-in supports quite a few fields

MarkKoz (approved):

jb3 (approved):

Issue #18, opened 2022-01-31
Install and configure Podman on all hosts


jb3:
- [x] Install Podman on all hosts
- [ ] Configure Podman to assign pods IPs on the Wireguard subnet allocated to each host
- [ ] Configure Podman services to use host cgroup and run managed by systemd
- [ ] Setup monitoring for Podman
- [ ] Talk more about how we want DNS to work

jb3:
Great work @jchristgit!

jchristgit:
Regarding "Configure Podman services to use host cgroup and run managed by systemd":

```
[Unit]
Description = Gobble the memory

[Service]
Environment=PODMAN_SYSTEMD_UNIT=%n
Type = forking

PIDFile=%t/%n-pid
RuntimeDirectory= %n
ExecStartPre = /bin/rm -f %t/%n-pid %t/%n-cid
ExecStart = /usr/bin/podman run \
  --cgroups=no-conmon \
  --cgroups=disabled \
  --log-driver=journald \
  --cidfile %t/%n-cid \
  --conmon-pidfile %t/%n-pid \
  --detach \
  --name gobbler \
  gobbler 
ExecStop = /usr/bin/podman stop --ignore --cidfile %t/%n-cid
ExecStopPost = /usr/bin/podman rm --ignore -f --cidfile %t/%n-cid
MemoryMax = 300M
MemoryLimit = 300M
```

jchristgit:
@jb3 I still don't get the podman IP part, but can you take care of it, or maybe we can set up a call and figure out how that's done?

jchristgit:
I am thrilled to announce that we have solved the issue of having to configure
a container engine on our hosts ourselves by deciding to configure a container
cluster management orchestration solution instead.

PR #19, opened 2022-02-01
Add podman role and improve playbook organization

jchristgit:
This PR adds a new podman role, see #18.

The playbook is merged into sections for each group of hosts that we
want to deploy to. To limit by role now, use tags, such as `-t
role::podman`.

### DIFF
--- a/roles/postgres/tasks/main.yml
+++ b/roles/postgres/tasks/main.yml
@@ -6,21 +6,29 @@
       - postgresql-contrib-{{ postgresql_version }}
       - libpq-dev
     state: present
+  tags:
+    - role::postgres
 
 - name: Check postgres is started and enabled on boot.
   service:
     name: '{{ postgresql_daemon }}'
     state: started
     enabled: true
+  tags:
+    - role::postgres
 
 - name: Add postgres users.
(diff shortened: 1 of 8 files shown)


roles/postgres/tasks/main.yml
```diff
@@ -6,21 +6,29 @@
       - postgresql-contrib-{{ postgresql_version }}
       - libpq-dev
     state: present
+  tags:
+    - role::podman
```
jb3, line 10: All the tags in this file are wrong

jb3 (changes requested):

roles/postgres/tasks/main.yml
jchristgit: Hello Joe!

I was heavily honoured to receive a review from this on this matter today. I know that time is short for everyone here, especially you, and I am glad that you managed to grant me three minutes of your free time to review this pull request.

Rest assured that I have addressed your review comment to fullest correctness by running a so-called "search-and-replace" via my text editor "neovim" by using `:%s/podman/postgres/g` in this file, `roles/postgres/tasks/main.yml`. Therefore, I consider this matter closed.

jb3 (approved):
Johannes, thanks.

PR #51, opened 2022-02-16
Add nginx deployment

### LINKED ISSUE #22: NGINX reverse proxy & certificate management deployment

jchristgit:
Includes documented roles for:
- installing nginx & configuring handlers
- installing the mTLS certificate for Cloudflare
- installing firewall rules

They are kept separate for now, for composability.

Closes #22.


roles/nginx-ufw/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: allow https traffic through the firewall
+  ufw:
+    app: WWW Secure
+    rule: allow
+    comment: nginx web server
```
ChrisLovering, line 6: So long as Nginx is installed before this is ran (which it currently is), this suggestion would work and be more explicit.
```suggestion
    app: Nginx HTTPS
    rule: allow
```

ChrisLovering (approved):
Just one small comment, everything else seems great 👌

MarkKoz:
Should this just be dropped if we're going to be creating separate roles anyway? https://github.com/python-discord/infra/blob/main/roles/ufw/tasks/main.yml#L33-L37

jchristgit:
> Should this just be dropped if we're going to be creating separate roles anyway? https://github.com/python-discord/infra/blob/main/roles/ufw/tasks/main.yml#L33-L37

Hmm, I'm 50/50 of it. On one hand it's nice to have a common base spot to configure this stuff, on the other hand, your configurations for services start drifting around in the repository. For instance - in your postgresql configuration you add the site database, then in your UFW configuration you add this port, .... Does that make sense?

MarkKoz:
> on the other hand, your configurations for services start drifting around in the repository. For instance - in your postgresql configuration you add the site database, then in your UFW configuration you add this port, .... Does that make sense?

Sort of. Why can't the same be said for this PR's approach where the there is a separate role for nginx ufw rather than a single nginx role for all nginx things?

jchristgit:
> Sort of. Why can't the same be said for this PR's approach where the there is a separate role for nginx ufw rather than a single nginx role for all nginx things?

Because NGINX and UFW are independent services, and the NGINX role does not perform any configuration of NGINX sites by itself.

MarkKoz:
> Because NGINX and UFW are independent services, and the NGINX role does not perform any configuration of NGINX sites by itself.

Isn't that the case for every service that needs a UFW config? I'm not following how creating a separate "<service>-ufw" role for each service is any better regarding "drifting around in the repository" than using a centralised location.

Issue #54, opened 2022-02-20
Setup firewall within VPN


jchristgit:
We need to allow our services to reach other within the VPN network. For instance, #23 needs access to port 9100 to scrape node-exporter, and I assume more hosts will be added in the future.

jchristgit:
- Allow all within the VPN network
- We can deny individual connections later as needed

MateusRomao:
Wich firewall is running on this infrastructure?

PR #55, opened 2022-02-20
Deploy Prometheus & node-exporter

jchristgit:
To start off, we are only scraping Prometheus itself and node-exporter.

image

This needs some firewall changes (separately), see #54.

### DIFF
--- a/host_vars/ritchie/prometheus.yml
+++ b/host_vars/ritchie/prometheus.yml
@@ -1,0 +1,35 @@
+---
+prometheus_configuration:
+  global:
+    scrape_interval: 15s  # Set the scrape interval to every 15 seconds. Default is every 1 minute.
+    evaluation_interval: 15s  # Evaluate rules every 15 seconds. The default is every 1 minute.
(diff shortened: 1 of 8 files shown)


host_vars/ritchie/prometheus.yml
```diff
@@ -0,0 +1,35 @@
...
+  alerting:
+    alertmanagers:
+      - static_configs:
+          - targets: []
+
+  rule_files:
+    # - "first_rules.yml"
+    # - "second_rules.yml"
+
+  scrape_configs:
```
ChrisLovering, line 18: Would this be where we add other services that need to be scraped?

When we get the VPN up and running, is this where we'd add the services running on other machines too?

jchristgit: Yes

ChrisLovering (approved):
Looks fine to me, I still haven't got personal infra setup to test it though.

jchristgit:
Thanks!

PR #56, opened 2022-02-23
Configure GeoIP for moon phase support

jchristgit:
In order to add moon phase support on the dark theme picker later, we
need to configure the GeoIP module included with nginx.

On Debian, the `nginx` package that we install installs `nginx-core`,
which in turn installs the GeoIP module and even a GeoIP country
database for us.

### DIFF
--- a/playbook.yml
+++ b/playbook.yml
@@ -16,6 +16,7 @@
   hosts: nginx
   roles:
     - nginx
+    - nginx-geoip
     - nginx-ufw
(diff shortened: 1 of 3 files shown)


ChrisLovering (approved):
This is vital in supporting the infrastructure here at Python Discord. I approve.

jchristgit:
Thanks!

jchristgit:
"Good move!" - Joe Banks

Issue #26, opened 2022-02-15
h deployment


jchristgit:
Our most important backing service, this needs to be set up as soon as possible due to the sheer amount of usefulness encapsulated in such a great deployment service.

This may be a good task to jump into the deployment setup with Ansible, and once we have the previous issues done, we can admit somebody to the DevOps team to take care of it.

imsofi:
I will give my sincerest comment on this issue, for Cleiner.

jchristgit:
- Template a systemd service to run podman
  - No registry authentication is needed
  - Resource limits would be great
- Template NGINX configuration
  - Help from @jb3 and @jchristgit for questions 
  - Reload nginx on change

For the nginx deployment, see https://github.com/python-discord/infra/tree/main/roles/nginx-cloudflare-mtls for the mutual TLS configuration. @jb3 knows more on this topic.

PR #57, opened 2022-03-06
Initial h-asgi configuration

soupglasses:
This is a basic setup for an h-asgi deployment. Me and @MarkKoz got as far as we could without access to any staging environment.

We found 2 major issues with trying to deploy h-asgi in this way. First being a possible issue around conflicting iptables rules with ufw and podman. Where a published port cannot be port forwarded easily. See: https://stackoverflow.com/a/51741599
This should not be an issue if we use only reverse proxies, like nginx. But is something to keep in mind as a possible issue if we ever need a podman container be exposed directly.

Second issue was how to properly implement podman containers in ansible. We did find a lot around how to set up with quotas and memory limits. But it could lead to issues with OOM nuking the download process for podman repeatedly, to the point it corrupts podman to segfault at any command. We can use the `Startup*=` limits to work around this issue and only download images in a `ExecStartPre` stage.

Finally, @jchristgit the template you linked has two cgroup definitions. As i have understood you cant stack these, but id like your word on your reason why you had written it like that before removing it.

### DIFF
--- a/roles/h-asgi/tasks/main.yml
+++ b/roles/h-asgi/tasks/main.yml
@@ -6,8 +6,9 @@
     - container:
         name: h-asgi
         image: ghcr.io/vcokltfre/h:latest
-        mem_max: 25M
-        cpu_quota: 3%
         port: 8000:8000
+    - service:
+        # CPUQuota: 3%
+        # MemoryMax: 25M
   tags:
     - role::h-asgi
--- a/roles/h-asgi/tasks/main.yml
+++ b/roles/h-asgi/tasks/main.yml
@@ -1,0 +1,13 @@
+---
+- name: Install h-asgi systemd file
+  include_role:
+    name: podman
+  vars:
+    - container:
+        name: h-asgi
+        image: ghcr.io/vcokltfre/h:latest
+        mem_max: 25M
+        cpu_quota: 3%
+        port: 8000:8000
+  tags:
+    - role::h-asgi
--- a/roles/podman/handlers/main.yml
+++ b/roles/podman/handlers/main.yml
@@ -1,0 +1,4 @@
+---
+- name: reload systemd
+  systemd:
+    daemon_reload: true
--- a/roles/podman/tasks/main.yml
+++ b/roles/podman/tasks/main.yml
@@ -1,7 +1,27 @@
 ---
-- name: install podman
+- name: Ensure package is installed
   package:
     name: podman
     state: present
+  tags:
+    - role::podman
+
+- name: Ensure systemd service is present
+  template:
+    src: templates/container-.service.j2
+    dest: "/etc/systemd/system/container-{{ container.name }}.service"
+    owner: root
+    group: root
+    mode: "0644"
+  notify:
+    - reload systemd
+  tags:
+    - role::podman
+
+- name: Ensure the systemd service is started and enabled
+  systemd:
+    name: "container-{{ container.name }}.service"
+    state: started
+    enabled: true
   tags:
     - role::podman
--- a/roles/podman/templates/container-.service.j2
+++ b/roles/podman/templates/container-.service.j2
@@ -1,0 +1,42 @@
+# vi: ft=systemd
+[Unit]
+Description = Podman container service for {{ container.name }}
+Wants=network-online.target
+After=network-online.target
+RequiresMountsFor=/var/run/container/storage
+
+[Service]
+Environment=PODMAN_SYSTEMD_UNIT=%n
+Type = forking
+
+PIDFile=%t/%n-pid
+Restart=on-failure
+RuntimeDirectory= %n
+
+ExecStartPre = /bin/rm -f %t/%n-pid %t/%n-cid
+# QUESTION: Why two cgroups? This was taken from:
+# https://github.com/python-discord/infra/issues/18#issuecomment-1027307623
+# TODO: make publish can be looped over to get multiple
+# ports opened.
+# TODO: look into an alternative for publishing ports
+# to the host adapter. Had issues with ufw conflicting when
+# using --publish directly. See: https://stackoverflow.com/a/51741599
+ExecStart = /usr/bin/podman run \
+  --cgroups=no-conmon \
+  --cgroups=disabled \
+  --log-driver=journald \
+  --cidfile %t/%n-cid \
+  --conmon-pidfile %t/%n-pid \
+  --detach \
+  --name {{ container.name }} \
+  --publish {{ container.port }} \
+  {{ container.image }}
+ExecStop = /usr/bin/podman stop --ignore --cidfile %t/%n-cid
+ExecStopPost = /usr/bin/podman rm --ignore -f --cidfile %t/%n-cid
+
+# TODO: Low values make the OOM mad when initially downloading a package.
+# Downloading packages should likely be done at a different place
+# or as a ExecStartPre with higher limits on CPUQuota and MemoryMax.
+
+#CPUQuota = {{ container.cpu_quota }}
+#MemoryMax = {{ container.mem_max }}
--- a/roles/podman/templates/container-.service.j2
+++ b/roles/podman/templates/container-.service.j2
@@ -41,5 +41,6 @@ ExecStopPost = /usr/bin/podman rm --ignore -f --cidfile %t/%n-cid
 # Downloading packages should likely be done at a different place
 # or as a ExecStartPre with higher limits on CPUQuota and MemoryMax.
 
-#CPUQuota = {{ container.cpu_quota }}
-#MemoryMax = {{ container.mem_max }}
+{% for key, value in (service | default({}) or {}).items() -%}
+{{ key }}={{ value }}
+{% endfor -%}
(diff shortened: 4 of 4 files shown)


imsofi:
This PR relates to issue #26

roles/podman/tasks/main.yml
```diff
@@ -1,7 +1,27 @@
...
+  tags:
+    - role::podman
+
+- name: Ensure the systemd service is started and enabled
+  systemd:
+    name: "container-{{ container.name }}.service"
+    state: started
+    enabled: true
+  tags:
+    - role::podman
```
jchristgit, line 27: This should probably be part of a different task in here, otherwise it will try to do this everytime we run the role, which also gets slow for many services. If we have the `h-asgi` role depend on `podman`, and `podman`'s `main` tasks only install podman, then once we have multiple roles that depend on `podman`, they will only run the role the first time they need it. So my recommendation would be:

- `h-asgi` gets a dependency to `podman` in `meta/main.yml`
- `podman` gets a new task e.g. `setup-systemd-container.yml`
- the ``h-asgi`` tasks should include tasks from somewhere else

roles/podman/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: reload systemd
+  systemd:
+    daemon_reload: true
```
jchristgit, line 4: For sake of correctness, this should probably be in a role called `systemd` with just this handler that's depended upon, otherwise we may end up duplicating this everywhere.

roles/h-asgi/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
...
+- name: Install h-asgi systemd file
+  include_role:
+    name: podman
+  vars:
+    - container:
+        name: h-asgi
+        image: ghcr.io/vcokltfre/h:latest
+        mem_max: 25M
+        cpu_quota: 3%
+        port: 8000:8000
```
jchristgit, line 11: Wait, how does this work with a list..??

roles/podman/templates/container-.service.j2
```diff
@@ -0,0 +1,42 @@
+# vi: ft=systemd
```
jchristgit, line 1: Absolute winner. ABSOLUTE winner.

roles/podman/templates/container-.service.j2
```diff
@@ -0,0 +1,42 @@
...
+[Service]
+Environment=PODMAN_SYSTEMD_UNIT=%n
+Type = forking
+
+PIDFile=%t/%n-pid
+Restart=on-failure
+RuntimeDirectory= %n
+
+ExecStartPre = /bin/rm -f %t/%n-pid %t/%n-cid
+# QUESTION: Why two cgroups? This was taken from:
```
jchristgit, line 17: I don't know either. I think only the second one is used though.

jchristgit (commented):
I see this as an absolute win. I left some Ansible suggestions:tm:

roles/h-asgi/tasks/main.yml
imsofi: Honestly, i have no idea. But it does :D

roles/podman/templates/container-.service.j2
```diff
@@ -22,7 +22,6 @@ ExecStartPre=/bin/rm -f %t/%n-pid %t/%n-cid
 # to the host adapter. Had issues with ufw conflicting when
 # using --publish directly. See: https://stackoverflow.com/a/51741599
 ExecStart=/usr/bin/podman run \
-  --cgroups=no-conmon \
```
imsofi: We should probably use the no-conmon option rather than completely disable cgroups. As it is the default option when generating a new podman service file with --new. Maybe @jchristgit remembers why there is a disabled on the cgroup here?

https://docs.podman.io/en/latest/markdown/podman-generate-systemd.1.html

MarkKoz: In #18, one of the goals is

> Configure Podman services to use host cgroup and run managed by systemd

According to @jchristgit, we want to use systemd for resource limits (meaning we don't need pdoman to create cgroups) for "ease of getting metrics."

roles/podman/templates/container-.service.j2
```diff
@@ -0,0 +1,46 @@
...
+
+PIDFile=%t/%n-pid
+Restart=on-failure
+RuntimeDirectory=%n
+
+ExecStartPre=/bin/rm -f %t/%n-pid %t/%n-cid
+# QUESTION: Why two cgroups? This was taken from:
+# https://github.com/python-discord/infra/issues/18#issuecomment-1027307623
+# TODO: make publish can be looped over to get multiple
+# ports opened.
```
MarkKoz, line 20: @imsofi Do you think this is resolved now that arbitrary options can be passed to podman? I imagine most services will only need to publish one port, so there isn't much benefit to the small convenience of supporting a list of ports directly.

roles/podman/templates/container-.service.j2
```diff
@@ -41,5 +41,6 @@ ExecStopPost = /usr/bin/podman rm --ignore -f --cidfile %t/%n-cid
 # Downloading packages should likely be done at a different place
 # or as a ExecStartPre with higher limits on CPUQuota and MemoryMax.
 
-#CPUQuota = {{ container.cpu_quota }}
-#MemoryMax = {{ container.mem_max }}
+{% for key, value in (service | default({}) or {}).items() -%}
+{{ key }}={{ value }}
+{% endfor -%}
```
imsofi, line 46: While a good idea for flexibility, it does pushes the complexity up the stack. Leading to a more fragmented configuration for all the roles that rely on podman. Where it makes it harder to spot mis-configured services, or force specific requirements on all podman services (for example that every service would need to define a value for CPUQuota or MemoryMax limits).

MarkKoz: I don't see why we can't have both.

roles/podman/templates/container-.service.j2
imsofi: No, arbitary arguments does not help adding extra options to the start command im afraid.

Using arbitrary options to simplify the complexity in our role is counter productive. As we are essentially pushing that complexity upwards. Where now everyone using the podman role now also needs to know how to configure systemd, leading to a very over-complicated interface to set up podman containers.

roles/podman/templates/container-.service.j2
imsofi: Im against it because of how much complexity it adds to the interface we are creating. And if we run without arbitrary key=value configuration, it makes it much more desirable to put the change you want in the podman role directly, rather than each podman role having to have several arbitrary key value settings copy-pasted and updated across the repo.

roles/podman/templates/container-.service.j2
MarkKoz: It's not using systemd options; it's https://github.com/python-discord/infra/pull/57/commits/3a826423014cf7c006feb05d74cf456a8abdb200

There is some merit to what you are saying, but I also think it's not unreasonable to expect some options to be known. I think it's a fair trade off for more flexibility. It seems tedious to maintain a template that needs to cover all options we need. Furthermore, because now instead of being able to leverage existing knowledge of podman/systemd, one has to study our specific template to see how to set an option.

roles/podman/templates/container-.service.j2
imsofi: Oh yeah, that makes sense. 😄

roles/podman/templates/container-.service.j2
```diff
@@ -35,7 +35,8 @@ ExecStart = /usr/bin/podman run \
   {% endfor -%}
   {{ container.image }}
 ExecStop = /usr/bin/podman stop --ignore --cidfile %t/%n-cid
-ExecStopPost = /usr/bin/podman rm --ignore -f --cidfile %t/%n-cid
+ExecStopPost=/usr/bin/test -f %t/%n-cid
+ExecStopPost=/usr/bin/podman rm --ignore -f --cidfile %t/%n-cid
```
imsofi, line 38: This is a nice idea :D

MarkKoz:
There is no `Startup*` equivalent for memory limits. I'm not sure if that means the memory limit applies to all stages or only to `ExecStart`.

imsofi:
> There is no `Startup*` equivalent for memory limits. I'm not sure if that means the memory limit applies to all stages or only to `ExecStart`.

Currently im looking into separating the downloading of the podman container, as it is not technically needed in the systemd file.

roles/podman/tasks/main.yml
imsofi: I think i understand your general idea, but I am still a bit unsure what you are proposing. Could we take a call when you have time and talk over this?

MarkKoz:
> > There is no `Startup*` equivalent for memory limits. I'm not sure if that means the memory limit applies to all stages or only to `ExecStart`.
> 
> Currently im looking into separating the downloading of the podman container, as it is not technically needed in the systemd file.

In that case, probably good to use `--pull=never` for `podman run`.

roles/podman/tasks/main.yml
jchristgit: If we have these tasks in the base role like this, every dependency to the `podman` role with different `vars` will re-run the role again and thus also run all the tasks in here, such as installing podman itself. We should split the configuration into a separate task file, that is then included when we want to configure a container.

jchristgit: We can also talk about this in a meeting, though. I'll message the channel shortly

PR #58, opened 2022-03-07
install Prometheus postgres exporter role

### LINKED ISSUE #25: PostgreSQL exporter deployment
"Sure!" - Chris

ChrisLovering:
Closes #25

### DIFF
--- a/roles/postgres/vars/main.yml
+++ b/roles/postgres/vars/main.yml
@@ -2,6 +2,8 @@ postgresql_version: "13"
 postgresql_daemon: "postgresql@{{ postgresql_version }}-main"
 postgres_user: "postgres"
 
-postgresql_users: []
+postgresql_users:
+  - name: prometheus
+    login_unix_socket: /run/postgresql/
 
 postgresql_databases: []
--- a/roles/prometheus-postgres-exporter/handlers/main.yml
+++ b/roles/prometheus-postgres-exporter/handlers/main.yml
@@ -1,8 +1,6 @@
 ---
 
 - name: restart prometheus-postgres-exporter
-  become: true
   service:
-    daemon_reload: true
     name: prometheus-postgres-exporter.service
     state: restarted
--- a/roles/prometheus-postgres-exporter/handlers/main.yml
+++ b/roles/prometheus-postgres-exporter/handlers/main.yml
@@ -1,0 +1,8 @@
+---
+
+- name: restart prometheus-postgres-exporter
+  become: true
+  service:
+    daemon_reload: true
+    name: prometheus-postgres-exporter.service
+    state: restarted
--- a/playbook.yml
+++ b/playbook.yml
@@ -29,3 +29,4 @@
   hosts: lovelace
   roles:
     - postgres
+    - prometheus-postgres-exporter
--- a/roles/prometheus-postgres-exporter/meta/main.yml
+++ b/roles/prometheus-postgres-exporter/meta/main.yml
@@ -1,0 +1,4 @@
+---
+
+dependencies:
+  - postgres
--- a/roles/prometheus-postgres-exporter/tasks/main.yml
+++ b/roles/prometheus-postgres-exporter/tasks/main.yml
@@ -1,0 +1,28 @@
+---
+- name: install prometheus-postgres-exporter
+  package:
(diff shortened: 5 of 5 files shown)


roles/postgres/vars/main.yml
```diff
@@ -2,6 +2,8 @@ postgresql_version: "13"
 postgresql_daemon: "postgresql@{{ postgresql_version }}-main"
 postgres_user: "postgres"
 
-postgresql_users: []
+postgresql_users:
+  - name: prometheus
+    login_unix_socket: /ran/postgresql/
```
MarkKoz, line 7:
```suggestion
    login_unix_socket: /run/postgresql/
```

MarkKoz (changes requested):

roles/postgres/vars/main.yml
ChrisLovering: Thanks mark https://github.com/python-discord/infra/pull/58/commits/7051c920e0216841f9ab9f25c7a30f2b8dd013c7

roles/prometheus-postgres-exporter/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+- name: restart prometheus-postgres-exporter
+  become: true
+  systemd:
+    daemon_reload: true
+    name: prometheus-postgres-exporter.service
+    state: restarted
```
MarkKoz, line 8: Can this instead use the `service` module which is a bit more generic, or is the a good reason to be using `systemd` here?

ChrisLovering: No, I don't believe there is a reason to use systemd here over service.

ChrisLovering: https://github.com/python-discord/infra/pull/58/commits/075e7d8b15646bdf62d3b04f3654b81a4f3b20e9

roles/prometheus-postgres-exporter/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+- name: restart prometheus-postgres-exporter
+  become: true
```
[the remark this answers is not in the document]
ChrisLovering: Ah yea, become is already yes in the config, so I do not believe this would be necessary.

roles/prometheus-postgres-exporter/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
...
+  service:
+    daemon_reload: true
```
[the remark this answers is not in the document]
ChrisLovering: Ah, it seems it doesn't so systemd would be needed to use this.

Talking with volcyy, it was needed to ensure that the service picked up the new config we set in the previous step.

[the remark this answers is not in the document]
jchristgit: For the systemd role, may be personal preference, but it may be nicer to have a `systemd` role that this role depends on, in that role just have a handler that does `daemon_reload`, otherwise we will start duplicating this across roles a lot.

roles/prometheus-postgres-exporter/handlers/main.yml
jchristgit: 🍣🍣🍣🍣🍣

roles/prometheus-postgres-exporter/handlers/main.yml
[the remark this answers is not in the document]
ChrisLovering: https://github.com/python-discord/infra/pull/58/commits/9a0e0321b55fd3ef323de7dfffeea3734c14bde1

roles/prometheus-postgres-exporter/handlers/main.yml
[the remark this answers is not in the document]
ChrisLovering: I've removed daemon_reload in this commit https://github.com/python-
discord/infra/pull/58/commits/9a0e0321b55fd3ef323de7dfffeea3734c14bde1

I don't quite understand your suggestion though volcyy, are you saying to add a new roll that just does the daemon_reload and the add that as a dependency in the meta of this role?

PR #60, opened 2022-03-14
Add certbot role

jb3:
Add a certbot role that generates a certificate on the first host in the
NGINX group and then deploys it to all other NGINX hosts. As of now we
generate wildcard certs for pythondiscord.com and pydis.wtf.

A unique SSH key is generated for each replica host which is restricted
for security purposes. A deploy hook is installed to push renewals to
other hosts.


jb3:
Made draft until I figure out how to reload NGINX on renewals.

PR #66, opened 2022-03-15
Default NGINX server

jb3:
When a NGINX node is reached with no matching server block, we will redirect to a random site from a list we build. Contributions are accepted.

### DIFF
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -1,7 +1,21 @@
 ---
-- name: Install nginx
+- name: Install NGINX & modules
   package:
-    name: nginx
+    name:
+      - nginx
+      - libnginx-mod-http-lua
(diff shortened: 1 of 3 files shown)


ChrisLovering (approved):
thanks

roles/nginx/tasks/main.yml
```diff
@@ -5,3 +5,23 @@
...
   tags:
     - role::nginx
+
+- name: Install NGINX modules
+  package:
+    name:
+      - libnginx-mod-http-lua
+    state: present
+  tags:
+    - role::nginx
```
jchristgit, line 15: Please install this in the parent task.

jchristgit (changes requested):

jchristgit (approved):
Great change.

PR #70, opened 2022-10-12
Delete roles we won't use with self-hosted k8s

ChrisLovering:
This deletes the following:

- auditbeat
- certbot
- elasticsearch
- filebeat
- jumpcloud
- kibana (+kibana-nginx)
- nginx (+nginx-cloudflare-mtls, nginx-geoip, nginx-ufw)
- packetbeat
- wireguard


jchristgit:
Legend!

PR #20, opened 2022-02-10
Add PostUp directive for routing wg subnet

jb3:
This PR adds a PostUp directive to the Wireguard config of all nodes to add the allocated /16 to the local routing table.

This means that hosts:
- Are aware that packets destined for these IPs should be captured by the Kernel, not ignored
- Allow services to bind to any IP address within their allocated /16

This does *not* include a PostDown directive since we do not want to lose the local routing should the Wireguard session go down, services should still be able to bind to relevant IPs even without Wireguard to prevent SPOF.

PR #105, opened 2023-06-21
Anti k8s infra

ChrisLovering:


roles/wireguard/templates/wg0.conf.j2
```diff
@@ -0,0 +1,25 @@
+# Configuration managed by Ansible
+[Interface]
+Address = {{ wireguard_subnet }}
+ListenPort = {{ wireguard_port }}
+PrivateKey = {{ wg_priv_key['content'] | b64decode | trim }}
+
+PostUp = ip route add local {{ wireguard_subnet }} dev eth0
```
jchristgit, line 7: Sure this is needed? Shouldn't `AllowedIPs` do this for us?

ChrisLovering: Haven't looked into it. This was specifically added in a separate PR here https://github.com/python-discord/infra/pull/20 `AllowedIPs` existed prior to this PR too.

roles/certbot/files/rsync.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Start the rsync server and perform the transfer
+rrsync -wo /etc/letsencrypt/live
```
jchristgit, line 4: I don't fully understand this. Won't this hang the certbot hook forever? Do we not want this to push to the hosts instead of waiting them to pull?

jchristgit: like, I understand that rsync is started below in the hook - but do we actually have a guarantee that they will all be ran at the same time? If not, do we need the rsync server or can we just rely on rsync to start it over e.g. `-e ssh`?

ChrisLovering: yea, we likely don't need the `rsync.sh` script at all. Not enreily sure what the nginx restart is for, but if it's needed, we could move it to a handler

ChrisLovering: https://github.com/python-discord/infra/pull/105/commits/8248baf78f82bee45958583f570104506f82face

PR #129, opened 2023-08-22
Snowflake proxy deployment

jchristgit:
I'm opening this as a draft because I want to do some testing on my server with strong sandboxing of this service first.


ChrisLovering:
After discussing with Admins we've come to the conclusion that since netcup gave us these servers as a sponsorship for our infra, running something that isn't our infra seemed wrong.

So while we think it's an important service, it doesn't seem appropriate to run it on this infra.

PR #164, opened 2023-12-05
Grafana role attribute path

ChrisLovering:
This maps github teams to roles on grafana, now that manually updating roles for users from external auth providers isn't possible


ChrisLovering:
Just need to hold off until directors make the directors team on Github before merging this

kubernetes/namespaces/default/grafana/configmap.yaml
```diff
@@ -13,7 +13,8 @@ data:
   GF_AUTH_GITHUB_ENABLED: "true"
   GF_AUTH_GITHUB_SCOPES: "user:email,read:org"
   # IDs can be retrieved via `gh api orgs/python-discord/teams`.
-  GF_AUTH_GITHUB_TEAM_IDS: "2638565,3854739,3114246"
+  GF_AUTH_GITHUB_TEAM_IDS: "2638565,3854739,3114246,7361120"
+  GF_AUTH_GITHUB_ROLE_ATTRIBUTE_PATH: "contains(groups[*], ['@python-discord/directors', '@python-discord/sudo-devops']) && 'Admin' || contains(groups[*], '@python-discord/admins') && 'Editor' || 'Viewer'"
```
jchristgit, line 17: What syntax is this?

Intuitively I would say that `|| 'String'` would always be true, because a non-empty string is a truthy value. But this probably doesn't work that way?

ChrisLovering: It's [JMESPath](https://jmespath.org/), examples are here https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/github/#role-mapping-examples looks weird to me too.

jchristgit: ahh, I understand. Ok, this is super cursed, but it looks like it will work

PR #172, opened 2024-01-08
Move non-secret secrets from secrets.yaml files to config maps

ChrisLovering:
this has not been applied


kubernetes/namespaces/default/site/secrets.yaml
ChrisLovering, line 1: This diff is just removing two unneeded new line chars

ChrisLovering:
merged & applied

PR #181, opened 2024-02-10
Script to create an x509 user config

ChrisLovering:
This script also updates a cluster role mapping to give the user correct permissions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ vault_passwords
 venv
 .cache/
 .vscode/
+*.config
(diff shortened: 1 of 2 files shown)


shtlrs:
I was trying to test this, the first thing I notice is the the script doesn't have any output, so as an executer I don't know what has succeeded/failed.

So when I tried to debug by hand, and list all the `CSR`s, I don't find any, which tells me that something went wrong while executing the script but I don't know what is.

Maybe you could also include `stderr` when executing your commands, and check it the run resulted in any errors, and if that's the case, log the error & exit early ?

shtlrs:
As I read through the script, the CSR doesn't exist since it's being deleted by the script, so that's normal.

Currently, it's erroring at `f.write(ROLE_BINDING_PATCH_TEMPLATE.format(user=user))` with a `KeyError`, and I think that's because you've already formatted the string at the beginning.

shtlrs (approved):

PR #195, opened 2024-03-26
Make Kubernetes user creation script Linux friendly

jb3:
Set `shell=True` to make the user generation script work on POSIX systems.

Additionally, change some quoting when fetching the Kubernetes CA to be POSIX friendly.

This needs to be tested on Windows to ensure the backwards compatibility is not affected. Due to the nature of how RBAC works, this is safe to run on your existing user account and will not remove any of your accesses/invalidate your tokens if you do want to test it.


ChrisLovering (approved):
Works on windows too. we love cross platform

PR #196, opened 2024-03-31
Support port pretty names as well as numeric ports for Kubernetes SD

jb3:
With this PR, we add support for the `prometheus.io/port` annotation to support both numeric ports as well as the pretty names of container ports.

We were experiencing an issue where non-metric endpoints were being collected into service discovery because Prometheus was seeing the presence of the scrape config but could not handle textual ports, hence it was selecting all ports as SD targets.

I have split the service discovery into two pools, `kubernetes-pods-port-integers` and `kubernetes-pods-port-numbers`, which look for numeric ports (and filter out text), and textual ports (and filter out non-matching ports) respectively.

This allows us to either put the pretty containerPort name under `prometheus.io/port` or the numeric text (right now, Loki uses textual ports and cert-manager uses numeric ones).


jchristgit (approved):
Nice, I'm a fan!

PR #198, opened 2024-03-31
More Prometheus Alerts

jb3:
- **Add some new Django alerts through Prometheus**
- **Add etcd alerts for increase in failed requests**
- **Add alerts for container OOMs**
- **Add a sanity alert for when a node becomes unschedulable**


kubernetes/namespaces/monitoring/alerts/alerts.d/memory.yaml
```diff
@@ -10,3 +10,12 @@ groups:
...
       summary: "Node {{ $labels.kubernetes_node }} has RAM usage >80% for 5 minutes"
       description: 'RAM usage is currently {{ $value | humanizePercentage }} on {{ $labels.kubernetes_node }}'
+
+  - alert: ContainerOOMEvent
+    expr: container_oom_events_total{pod=~".+", container_name!="POD", image!=""} > 0
+    for: 30s
+    labels:
+      severity: page
+    annotations:
+      summary: "Container {{ $labels.container_name }} in {{ $labels.pod }} has OOM killed"
```
jchristgit, line 20:
```suggestion
      summary: "Container {{ $labels.container_name }} in {{ $labels.pod }} was OOM killed"
```

kubernetes/namespaces/monitoring/alerts/alerts.d/etcd.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - name: etcd
+    rules:
+      - alert: EtcdErrorsSpike
+        expr: rate(etcd_request_error_total[5m]) > 0
+        for: 5m
+        labels:
+          severity: page
+        annotations:
+          summary: "Etcd request errors are spiking"
+          description: "Etcd request errors are spiking on {{ $labels.instance }} for {{ $labels.operation }}s to {{ $labels.type }}"
```
jchristgit, line 11: Can we do anything about these? It feels like a Kubernetes-internal thing that will drive one of us crazy at 3 AM when LKE does their production rollout.

kubernetes/namespaces/monitoring/alerts/alerts.d/nodes.yaml
```diff
@@ -41,9 +41,18 @@ groups:
...
       description: "Node {{ $labels.kubernetes_node }} has been unready for a long time"
+
+  - alert: KubernetesNodeCordoned
+    expr: kube_node_spec_unschedulable == 1
+    for: 1m
+    labels:
+      severity: page
+    annotations:
+      summary: Kubernetes node ({{ $labels.kubernetes_node }}) is cordoned
+      description: "Node {{ $labels.kubernetes_node }} has been cordoned"
```
jchristgit, line 58: Why is this an alert? Isn't it a manual action? If it's a manual action, shouldn't we bump up the `for` time to something like 1-2 hours to leave it as a "in case you forgot it" alert?

kubernetes/namespaces/monitoring/alerts/alerts.d/django.yaml
```diff
@@ -0,0 +1,20 @@
...
+        expr: increase(django_http_responses_total_by_status_total{status=~"5.."}[5m]) > 0
+        for: 5m
+        labels:
+          severity: page
+        annotations:
+          summary: "Django is experiencing 5xx errors"
+          description: "Django is experiencing 5xx errors on {{ $labels.namespace }}/{{ $labels.job }}"
+
+      - alert: DjangoLatencyHigh
+        expr: histogram_quantile(0.95, rate(django_http_requests_latency_seconds_by_view_method_bucket{view!="api:github-artifacts"}[5m])) > 1.0
```
jchristgit, line 14: I have a feeling that this will blow up in some cases on endpoint that receive 1 request every full moon and then that request is slow for some reason. Like, I think that we should maybe change this in a way that it will only do the monitoring on endpoints that receive a lot of requests. But I think for now this is fine.

jchristgit (changes requested):

kubernetes/namespaces/monitoring/alerts/alerts.d/etcd.yaml
jb3: This should not change as a result of anything LKE do, if we experience this it should be something that we can fix by recycling a node and it is often a precursor to wider network issues on a node.

kubernetes/namespaces/monitoring/alerts/alerts.d/nodes.yaml
jb3: Good point, I'll up it to 30 minutes. In theory, a node could become "unschedulable" through something that we don't do, so I think having it somewhat low is a good idea in case something else happens, but I'll up it to 30 minutes.

kubernetes/namespaces/monitoring/alerts/alerts.d/django.yaml
jb3: Probably yes, someone smarter than me could probably use `django_http_requests_total_by_view_transport_method_total` and validate there have been more than just one-shot requests to a view.

jchristgit (approved):
good

PR #206, opened 2024-04-10
Add a users role

jchristgit:
The new `pydis-users` role allows us to manage user accounts and move away from the root user setup script, eventually locking down SSH access to the root user.

Please add your users and SSH keys to this.


jchristgit:
Problem: ansible-lint throws up when encountering git-crypted files.
Solution: delete git-crypted files before running ansible-lint.

However our CI knows the vault password. I argue it's more secure this way, and we still lint our tasks, so it's fine.

ChrisLovering (approved):

PR #209, opened 2024-04-11
Configure SSH daemon options in unit dropin

jchristgit:
Disable password authentication and root logins and use a configuration file that is independent of the `sshd_config` that `apt` itself will modify on upgrades.

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -14,14 +14,21 @@
   tags:
     - role::common
 
-- name: Disable SSH password authentication
-  lineinfile:
-    dest: /etc/ssh/sshd_config
-    regexp: "^PasswordAuthentication"
-    line: "PasswordAuthentication no"
-    state: present
+- name: Configure SSH daemon options
+  ansible.builtin.copy:
+    content: |
+      # Ansible managed
+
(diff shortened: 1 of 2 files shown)


ansible/roles/common/tasks/main.yml
```diff
@@ -14,14 +14,21 @@
...
-    regexp: "^PasswordAuthentication"
-    line: "PasswordAuthentication no"
-    state: present
+- name: Configure SSH daemon options
+  ansible.builtin.copy:
+    content: |
+      # Ansible managed
+
+      PasswordAuthentication no
+      PermitRootLogin no
```
shtlrs, line 23: Is it the best option to remove root logins here ? (I'm just thinking out loud)
What I wonder is, if someone with root access needs to perform an adhoc action and is the only one available and either:
- using another laptop, so the ssh key isn't available
- has lost their ssh keys, for whatever reason and wants to gen a new pair and install it

ChrisLovering: this removes login via ssh for the user `root`. If you don't have your ssh key, then you won't be able to login as root over ssh anyway.

This does not block us from accessing root via the netcup recovery terminal if something is really needed.

ChrisLovering (approved):

PR #220, opened 2024-04-14
Use Ansible Vault for storing users

### LINKED ISSUE #211: Use Ansible Vault for pydis users
With yesterday's merge, we have integrated user management into our Ansible
deployment.

@jb3 raised the suggestion that we should use Ansible Vault for this. Since this
is natively integrated to Ansible, I agree it's preferable.

However, the current `git-crypt` workflow is very straightforward as `git-crypt`
is fast and the `.gitattributes` filters allow the repository to be
transparently handled locally. However, it seems this is still possible - minus
editing - using `.gitattributes`. See [this StackOverflow
post](https://stackoverflow.com/questions/29937195/how-to-diff-ansible-vault-changes).
With that implemented, we can switch this to Ansible vault.

jchristgit:
Closes #211.


ChrisLovering (approved):

jchristgit:
LGTM!

PR #224, opened 2024-04-15
Ensure linting CI runs when we update dependency versions

### LINKED ISSUE #223: Pin ansible-lint version used in CI
At the moment, our ansible-lint version found in
`.github/workflows/lint-ansible.yaml` is not pinned in any way. This means that
unexpected lint changes may pop up in unrelated pull requests and cause the
linter to throw up, which is not helpful.

We should use the standard [ansible-lint
package](https://pypi.org/project/ansible-lint) and use the Ansible Lint GitHub
Action as documented there.

ChrisLovering:
Closes #223


jchristgit (approved):
Thanks!

PR #231, opened 2024-04-27
OctoDNS

jb3:
This PR installs and configures [OctoDNS](https://github.com/octodns/octodns), which will allow us to configure and deploy all DNS changes through GitOps instead of having to log into Cloudflare.

### DIFF
--- a/.github/workflows/dns-deploy.yaml
+++ b/.github/workflows/dns-deploy.yaml
@@ -1,0 +1,27 @@
+name: Deploy DNS to providers
+
+on:
+  push:
+    branches:
+      - main
+jobs:
+  octodns-sync:
+    environment: production
+    name: Sync latest changes to DNS providers
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./dns
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - run: pip install -r requirements.txt
+      - uses: solvaholic/octodns-sync@main
+        with:
+          config_path: dns/production.yaml
+          doit: '--doit'
+        env:
+          CLOUDFLARE_TOKEN: ${{ secrets.CLOUDFLARE_TOKEN }}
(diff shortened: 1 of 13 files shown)


.github/workflows/dns.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.12'
+      - run: pip install -r requirements.txt
+      - uses: solvaholic/octodns-sync@main
+        id: generate-plan
+        with:
+          config_path: dns/production.yaml
+        env:
+          CLOUDFLARE_TOKEN: ${{ secrets.CLOUDFLARE_TOKEN }}
```
jchristgit, line 23: This begs the question: How secure is it? I suppose once again as secure as our branch restrictions?

jb3: In which sense?

It will not deploy any changes that remove over 30% of entries, and we obviously do not even run the plan on fork PRs.

We could try enforce it so DNS changes have to go through a PR process and have a plan generated, or tighten our branch restrictions.

jchristgit: My worry is just bad actors being able to tinker with our DNS. I think our staff are trustworthy, but with the current setup, over 150 people are able to tinker with our DNS, from my understanding:

image

Should we restrict these permissions?

jb3: I have just checked on this.

We enforce PRs for the main branch as well as restricting that only members of python-discord/devops can push to this branch, so it should be sufficiently safe.

jchristgit: What prevents changing the actions workflow from the branch such that it will deploy the changes on the branch? The token is available, or is the non-master token read-only?

jb3: Hmmm this is a good point, I can split it into read-only & write tokens for when I add the deployment flow.

ChrisLovering (approved):

jchristgit (approved):
Didn't review the DNS records but I assume you're not going to knock us offline as soon as it is merged.

PR #235, opened 2024-04-27
Stop alerting for slow GitHub webhook filter endpoint calls

jchristgit:
These are directly forwarded to GitHub with no time-consuming processing
done on the site. We would therefore be alerting for GitHub's slowness,
which is rather useless.

### DIFF
--- a/kubernetes/namespaces/monitoring/alerts/alerts.d/django.yaml
+++ b/kubernetes/namespaces/monitoring/alerts/alerts.d/django.yaml
@@ -11,7 +11,7 @@ groups:
           description: "Django is experiencing 5xx errors on {{ $labels.namespace }}/{{ $labels.job }}"
 
       - alert: DjangoLatencyElevated
-        expr: histogram_quantile(0.95, rate(django_http_requests_latency_seconds_by_view_method_bucket{view!="api:github-artifacts", view!="home:home", view!="content:tag"}[5m])) > 1.0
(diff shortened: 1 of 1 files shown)


kubernetes/namespaces/monitoring/alerts/alerts.d/django.yaml
```diff
@@ -11,7 +11,7 @@ groups:
           description: "Django is experiencing 5xx errors on {{ $labels.namespace }}/{{ $labels.job }}"
 
       - alert: DjangoLatencyElevated
-        expr: histogram_quantile(0.95, rate(django_http_requests_latency_seconds_by_view_method_bucket{view!="api:github-artifacts", view!="home:home", view!="content:tag"}[5m])) > 1.0
+        expr: histogram_quantile(0.95, rate(django_http_requests_latency_seconds_by_view_method_bucket{view!="api:github-artifacts", view!="api:github-webhook-filter", view!="home:home", view!="content:tag"}[5m])) > 1.0
```
shtlrs, line 14: I don't know how prometheus alerting works at all

But, isn't `github-webhook-filter` just the name you assign to that view in site's api ? The actual url contains the webhook id and webhook token as path params

Will it then be able to know how to ignore it still ?

Or maybe there's some mechanism i'm not aware of to how these views are labeled/managed

shtlrs (approved):
SEND IT

PR #257, opened 2024-04-29
Use nftables for firewalling

jchristgit:
nftables is the modern replacement for iptables, which ufw uses under
the hood. It allows us to specify firewall rules in a simple text file
(with as much or as little abstraction as we want) and is quick to
update and read. The text-file format allows more liberty with
commenting compared to UFW. The existing `ufw` role has been converted
to simply remove UFW. This has already been deployed on lovelace.


jchristgit:
Thanks!

PR #260, opened 2024-04-29
Update nftables role

jchristgit:
The new commit includes automatic validation of the `nft` configuration
to ensure that any deployed config is valid.


jchristgit:
Thanks!

PR #263, opened 2024-04-29
Add a couple of enhancements to the local testing guide.

shtlrs:
This does the following:
* Update debian box version used
* Add the missing ansible installation during the provisioning of VMs
* Fix the wrong synced folder paths in the docs
* Add more fixes/notes on what could go wrong
* Make the readme richer by setting up execution contexts/info

### DIFF
--- a/ansible/local_testing/README.md
+++ b/ansible/local_testing/README.md
@@ -1,27 +1,51 @@
 # Testing Locally
 
+This section contains the necessary steps in order to be able to setup virtual machines that mimic our netcup servers and be able to run the ansible roles/playbooks we have on them.
+
 ### Requirements
 
 - [Vagrant](https://developer.hashicorp.com/vagrant/docs/installation)
 - [VirtualBox](https://www.virtualbox.org/wiki/Downloads)
 
-## Get Started
 
-```shell
-vagrant up                  # This will take a while
-vagrant ssh                 # Get a shell, password=vagrant
+### Getting started
 
-# inside control VM
-/vagrant/scripts/push-keys  # Push the control VM's ssh key to all nodes
+1. Creating the virtual boxes
 
-# run ansible
-cd infra
-# if you're on Windows, the `infra` directory will be mounted with permissions set to 777
-# which ansible will reject as insecure
-# export the below environment variable to force anisble to accept the writable config
-# export ANSIBLE_CONFIG='/home/vagrant/infra/ansible.cfg'
+On your host machine, run the following:
+
+```bash
+cd ansible/local_testing
+vagrant up
+```
+
+This will create all the virtual boxes on which we will run our ansible playbooks.
+
+2. Push the control VM's ssh key to all nodes
+
+
+```bash
+vagrant ssh  # This will give you terminal access to the `control` VM, which we will use as the ansible control node.
+
+cd /home/vagrant/infra
+bash ./ansible/local_testing/scripts/push-keys
+```
+
+> If you're on windows, you might run into line ending issues when running the `push-keys` script.
+> To fix it, run this **while still inside the ssh session*: `sed -i 's/\r$//' /home/vagrant/infra/ansible/local_testing/scripts/push-keys`
+
+
+3. Run the ansible setup playbook
+
+```bash
+cd /home/vagrant/infra/ansible
 ansible-playbook playbook.yml --inventory local_testing/hosts.yaml --user vagrant
 ```
+
+> This will prompt you for the sudo password, whose value is `vagrant`.
+
+
+### Virtual machine IPs
 
 Below are the IPs of the VMs on the VirtualBox network
 ```yaml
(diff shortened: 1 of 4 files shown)


ansible/local_testing/README.md
```diff
@@ -1,28 +1,52 @@
...
+
+```bash
+vagrant ssh  # This will give you terminal access to the `control` VM, which we will use as the ansible control node.
+
+cd /home/vagrant/infra
+bash ./local_testing/scripts/push-keys
+```
+
+> If you're on windows, you might run into line ending issues when running the `push-keys` script.
+> To fix it, run this **while still inside the ssh session*: `sed -i 's/\r$//' /home/vagrant/infra/local_testing/scripts/push-keys`
```
jchristgit, line 35: I'm not on Windows, but shouldn't the file be checked in without `\r` line endings and thus not cause any issues?
Although it might still, because it's Windows. I'm not going to act like it makes sane decisions

ansible/local_testing/README.md
```diff
@@ -37,12 +61,25 @@ vms:
 
 # Fixes/Notes
 
+
+### Ansible cannot decrypt the files encrypted with ansible vault.
+
+* The `ansible\roles\certbot\vars\main.yaml` and `ansible\roles\pydis-users\vars\main.yaml` files have been encrypted with ansible vault due to their sensitive content.
```
jchristgit, line 67: Should we use forward slashes here? On both GitHub and any non-Windows system those are used, so you could also e.g. link to them by using forward slashes. I would find it better for consistency

ansible/local_testing/README.md
```diff
@@ -37,12 +61,25 @@ vms:
...
+
+* If you're not part of the people with access to the vault, you're going to have to either define your own variables or just exclude the appropriate roles.
```
jchristgit, line 69: Maybe we can make this a bit more concise:
```suggestion
* If you lack access to the vault, you're going to have to either define your own variables or exclude the appropriate roles.
```

ansible/local_testing/Vagrantfile
```diff
@@ -2,19 +2,20 @@
 # vi: set ft=ruby :
 
 Vagrant.configure("2") do |config|
-    config.vm.box = "bento/debian-11"
-    config.vm.box_version = "202212.11.0"
+    config.vm.box = "bento/debian-12.5"
```
jchristgit, line 5: Should we just pin to `bento/debian-12` to automatically get new patches? They shouldn't break anything, it's Debian after all

ansible/local_testing/Vagrantfile
```diff
@@ -2,19 +2,20 @@
...
+    config.vm.box_version = "202404.23.0"
     config.vm.provision "shell", inline: <<-SHELL
-        apt-get update
+        sudo apt-get update
         apt-get install -y python3 openssh-server
+        apt-get install -y python3 ansible
```
jchristgit, line 10: Let's use a single command:
```suggestion
        apt-get install -y python3 openssh-server ansible
```
Also I believe in the repository we currently install Ansible via poetry. Do you believe it's possible to also do that here?

ansible/local_testing/Vagrantfile
```diff
@@ -2,19 +2,20 @@
...
         systemctl enable ssh
     SHELL
 
     config.vm.define "control", primary: true do |control|
         control.vm.hostname = "control"
         control.vm.network "private_network", ip: "192.168.56.1",
             virtualbox__intnet: true
-        control.vm.synced_folder "../", "/home/vagrant/infra"
+        control.vm.synced_folder "../", "/home/vagrant/infra", type: "rsync"
```
jchristgit, line 18: Does the regular synced folder not work here?

jchristgit (changes requested):
Thanks!
I have some small thoughts and ideas but in general it looks good.

ansible/local_testing/README.md
shtlrs: That's the theory indeed, but when I tried  to push the SSH keys it started complaining :/

Should we maybe add it to .gitannotations ?

ansible/local_testing/README.md
shtlrs: Simple & elegant, granted

ansible/local_testing/README.md
shtlrs: Sounds good to me, I'll make the change

ansible/local_testing/Vagrantfile
shtlrs: It wasn't syncing the folders for me, I ended up with an emtpy /home/vagrant directory

ansible/local_testing/Vagrantfile
shtlrs: Yeah that sounds good, I'll make the change

ansible/local_testing/Vagrantfile
shtlrs: Having looked at the docs, it says vagrant chooses the best fitting sync type based on your setup.

I think it doesn't hurt to set an explicit sync type here better to rely on something we do not see nor have control over

jchristgit:
Hmm, I think that's worth a shot! If we can make the local setup easier
in any way we should.

ansible/local_testing/README.md
jchristgit: Hmm, I think that's worth a shot! If we can make the local setup easier
in any way we should.

ansible/local_testing/Vagrantfile
jchristgit: I think I'd still be in favour to use our regular Poetry version, because if we implement changes such as #264 and also in general the playbook is somewhat coupled to the Ansible version. I don't expect us to break compatibility much but I think to just prevent the possibility of breaking it for the Vagrant setup in the first place we should install it from Poetry. (It's also available as a Debian package)

ansible/local_testing/Vagrantfile
jchristgit: Sounds good to me. Strange that it doesn't work properly though. Perhaps once again a :sparkles: Windows speciality :fireworks:

jchristgit (approved):
:shipit::shipit::shipit::shipit::shipit:

jchristgit (approved):
Let's GO

PR #267, opened 2024-05-01
Depend on ansible-core instead of Ansible

jchristgit:
Allow for faster local installation by only installing what we need.


jchristgit:
Sorry for the updates, this is properly ready for review now.

shtlrs (changes requested):
Looks great ! 

Just one thing left to do due to the changes we've just merged for local testing: the vagrant file will need to be updated to install the custom galaxy collections.

jchristgit:
I updated the `Vagrantfile`. I can't test it locally but I'm relatively convinced it works like this.

shtlrs (approved):
Tested and works well, thanks !

image

PR #270, opened 2024-05-01
Remove old groups from Vagrant inventory

jchristgit:
These groups are no longer present in our proper inventory as we no
longer plan on selfhosting Kubernetes on the netcup nodes.


shtlrs (approved):
🥇

PR #272, opened 2024-05-01
Set up Prometheus alerting on Netcup

jchristgit:


ansible/roles/prometheus/defaults/main.yml
```diff
@@ -19,8 +19,7 @@ prometheus_configuration:
 
   # Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
   rule_files:
-    # - "first_rules.yml"
-    # - "second_rules.yml"
+    - "rules.yml"
```
shtlrs, line 22: Will ansible install prometheus in /etc/prometheus by default ?

I'm asking because i'm curious to how the file discovery works

jchristgit: Prometheus will use `/etc/prometheus/prometheus.yml` as its standard and default configuration file - Debian doesn't overwrite this. So the rule file will be loaded from `/etc/prometheus/rules.yml`

shtlrs (approved):
[Hit it](https://images-ext-1.discordapp.net/external/xctAOsBrkmsRz9H9qtxE2Cj7_TYv3gfWY3p1tw5Irbs/https/media.tenor.com/Zy2xtifOCC0AAAPo/keemxdaxtruth-lmao.mp4)

jchristgit:
Thanks!

PR #275, opened 2024-05-04
Configure default security limits

jchristgit:
The new limits allow each user to run a maximum of 100 processes by
default, allowing to manually raise this number to 200.

When a custom "pydis" group or similar is introduced, I plan to expand
this to also specify other limits to prevent user error from causing
problems on the system.

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -47,6 +47,21 @@
   tags:
     - role::common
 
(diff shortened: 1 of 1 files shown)


ChrisLovering (approved):
I am sad I can no longer fork bomb our own servers, but I will concede that this is a good change.

jchristgit:
Thanks!

PR #278, opened 2024-05-04
Configure Prometheus alerting for failed systemd units

jchristgit:
The two services that I would normally exclude are intentionally not
excluded right now to test out the alertmanager setup. If all goes well,
we should receive a notification on Discord.

### DIFF
--- a/ansible/roles/prometheus/tasks/main.yml
+++ b/ansible/roles/prometheus/tasks/main.yml
@@ -37,7 +37,7 @@
   copy:
     content: |
       # Ansible managed
-      {{ prometheus_rules | to_nice_yaml }}
+      {{ prometheus_rules }}
     dest: /etc/prometheus/rules.yml
     owner: prometheus
     group: prometheus
--- a/ansible/host_vars/lovelace/prometheus.yml
+++ b/ansible/host_vars/lovelace/prometheus.yml
@@ -73,3 +73,18 @@ prometheus_configuration:
         - source_labels: [__param_target]
           target_label: instance
         - target_label: __address__
+
+prometheus_rules: |
+  {% raw %}
+  groups:
+    - name: node
+      rules:
+        - alert: node/systemd-unit-failed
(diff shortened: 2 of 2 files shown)


ansible/roles/prometheus/defaults/main.yml
```diff
@@ -43,5 +43,17 @@ prometheus_configuration:
...
 
-prometheus_rules:
-  groups: []
+prometheus_rules: |
+  {% raw %}
+  groups:
+    - name: node
+      rules:
+        - alert: node/systemd-unit-failed
+          # expr: node_systemd_unit_state{state="failed", name!="openipmi.service", name!="nvmf-autoconnect.service"} != 0
```
shtlrs, line 52: Do you want to keep that commented out expression ?a

ansible/roles/prometheus/defaults/main.yml
```diff
@@ -43,5 +43,17 @@ prometheus_configuration:
       static_configs:
         - targets: ['localhost:9100']
 
-prometheus_rules:
-  groups: []
+prometheus_rules: |
```
shtlrs, line 46: The default config has been wiped out, so this should go to lovelace's hostvars

ansible/roles/prometheus/tasks/main.yml
```diff
@@ -37,7 +37,7 @@
   copy:
     content: |
       # Ansible managed
-      {{ prometheus_rules | to_nice_yaml }}
```
shtlrs, line 40: Yaml not nice, yaml grounded for eternity.

ansible/roles/prometheus/defaults/main.yml
jchristgit: Yes, it's intentional so we can just comment it back in and remove the other one once we've verified Alertmanager works.

ansible/roles/prometheus/defaults/main.yml
jchristgit: Done

shtlrs:
Thanks!!

shtlrs:
Thanks !!

PR #279, opened 2024-05-04
Install blackbox exporter as part of out monitoring stack

### LINKED ISSUE #218: Set up prometheus-blackbox-exporter in Ansible
Add a new role, named `prometheus-blackbox-exporter` (the
`prometheus-node-exporter` role can be taken as a guideline), and adjust our
playbook to install it on lovelace as part of the monitoring play.

Next up, adjust the Prometheus configuration (found in the Prometheus role
defaults) to scrape the blackbox exporter.

The `prometheus-blackbox-exporter` role should also allow configuration of the
blackbox exporter. For this, copy the configuration file from the current
distribution default, place it in the `templates` directory of the role, and
`template:` it over as part of the role. The exporter should be restarted on
changes. We might want to customize this such that an SSH access check can be
used.

shtlrs:
Closes #218

Prometheus' defaults are deleted since we rely on hostvars


ansible/roles/prometheus/defaults/main.yml
```diff
@@ -43,5 +43,41 @@ prometheus_configuration:
...
         - targets: ['localhost:9100']
 
+    - job_name: blackbox-ssh
+      metrics_path: /probe
+      params:
+        module: [ssh_banner]
+
+      static_configs:
+        - targets:
+            - lovelace.box.pydis.wtf
```
jchristgit, line 53: Wanna add turing here as well?

ansible/roles/prometheus/defaults/main.yml
```diff
@@ -43,5 +43,41 @@ prometheus_configuration:
...
+        - target_label: __address__
+          replacement: localhost:9115
+
+    - job_name: blackbox-http
+      metrics_path: /probe
+      params:
+        module: [http_2xx]
+      static_configs:
+        - targets:
+            - https://pydis.wtf/
```
jchristgit, line 74: We should also add https://cloud.native.is.fun.and.easy.pydis.wtf to the list

jchristgit (commented):
The Mind Toils Restlessly

ansible/roles/prometheus/defaults/main.yml
shtlrs: Done

ansible/roles/prometheus/defaults/main.yml
shtlrs: Granted

jchristgit (approved):
Looks good!

jchristgit:
Thanks!

PR #297, opened 2024-05-17
Add bitwarden postgres user & db config

shtlrs:


shtlrs:
This has been applied on lovelace

PR #303, opened 2024-05-19
Enforce SSL for remote PostgreSQL connections

jchristgit:


shtlrs (approved):
Dear Johannes,

We at Python Discord appreciate all the efforts you put into place to make things secure.

It is everything but unnoticeable how much effort and dedication you have to ensure no one eavesdrops on our services while they have a chat and sip coffe. 

All the way from enforcing SSL to our postgres database, to making sure that people are being good citizens by implementing the nftables rules to only allow them to connect over SSH, you never ceized to amaze us.

Thank you for the efforts, and here is a bunch of `Zimsterne` especially for you.

image

jb3:
Thanks!

PR #305, opened 2024-05-26
Generate a certificate for `prometheus.lovelace.box.pydis.wtf`

shtlrs:
This will be used to allow traffic to the prometheus instance on lovelace, enabling grafana to pull metrics.


jb3 (approved):

jb3:
Thanks!

PR #314, opened 2024-05-27
Fix AlertManager Discord instance formatting

jb3:
We made a change to include the instance in alerts sent to Discord, but
not all of our configured alerts send this field.

As a result, we would have incorrectly formatted alerts being sent
through to Discord which were tricky to read.

The format template has now been changed to only conditionally render
the instance label if it is present on a triggered alert.


ChrisLovering:
Thanks!

PR #327, opened 2024-05-31
Make issuing pg grants configurable

shtlrs:
This adds a task that allows issuing particular grants to specific roles in postgres


jb3 (dismissed):

jb3 (approved):
LOVE IT!

PR #331, opened 2024-06-01
[ansible] PostgreSQL exporter

jb3:
- Templates Prometheus config instead of using YAML to allow for more granular
  selection of scrape targets
- Enable the scraping of port 9187 for PostgreSQL exporter on all hosts in the
  `databases` group.


shtlrs (approved):
Thanks !!

PR #341, opened 2024-06-02
PostgreSQL mTLS + DevOps account

jb3:
Adds the necessary PostgreSQL configuration to allow the DevOps account to
authenticate with mTLS using short-lived certificates.

Additionally, add the necessary backing configuration to allow our other
services to connect with valid short-lived mTLS certificates which can be issued
by the Vault mTLS injector.


ansible/roles/postgres/vars/main/main.yml
```diff
@@ -51,6 +51,11 @@ postgres_users:
     roles:
       - pg_read_all_data
 
+  - name: devops
+    password: "{{ vault_postgres_user_passwords.grafana }}"
```
shtlrs, line 55:
```suggestion
    password: "{{ vault_postgres_user_passwords.devops}}"
```

ansible/roles/postgres/vars/main/main.yml
```diff
@@ -163,6 +178,8 @@ postgres_databases:
     owner: metabase
   - name: pixels
     owner: pixels
+  - name: devops
```
shtlrs, line 181: Why do you need a database for devops ?

ansible/roles/postgres/vars/main/main.yml
jb3: Good spot, fixed in force push.

ansible/roles/postgres/vars/main/main.yml
jb3: It's part of the configuration needed for mTLS as far as I am aware, the CN must match the database at least for the initial connection.

However, on top of that I think that given the account is a superuser connecting to a dummy zero-use database is a better idea than defaulting to something with more privilege/importance. I'd rather have people connect and then change database.

ansible/roles/postgres/tasks/main.yml
```diff
@@ -49,8 +50,14 @@
     insertafter: "# Put your actual configuration here"
     marker: "# {mark} ANSIBLE MANAGED HBA CONF BLOCK"
     block: |
+      # Manually configured HBA rules
       {% for rule in postgres_hba_rules %}
-      {{ rule.conn_type }}    {{ rule.database }}    {{ rule.user }}    {{ rule.address }}    {{ rule.method }}
+      {{ rule.conn_type }}    {{ rule.database }}    {{ rule.user }}    {{ rule.address }}    {{ rule.method }}   {{ rule['options'] | default('') }}
```
shtlrs, line 55: Just a nit pick, but we not keep on using dot notation to access attributes ? You have it in the `role_attr_flags` part as well

jb3: Fixed in force-push, I had a brain error moment re: using default.

ansible/roles/postgres/tasks/main.yml
```diff
@@ -101,3 +108,15 @@
     - role::postgres
   notify:
     - Restart the postgres service
+
+- name: Import PostgreSQL identity map (pg_ident.conf)
```
shtlrs, line 112: The docs say the following
> The pg_ident.conf file is read on start-up and when the main server process receives a SIGHUP signal. If you edit the file on an active system, you will need to signal the postmaster (using pg_ctl reload, calling the SQL function pg_reload_conf(), or using kill -HUP) to make it re-read the file.

Wouldn't this mean that we can just reload the service instead of restarting it here ?

shtlrs (approved):
Looks great, I just have one question but it's not necessary to implement

PR #346, opened 2024-06-06
Add Metricity manifest

jb3:
Copies the Metricity deployment manifest from the Metricity repo.

Once this PR has merged, I will update the metricity repo to use this manifest
on deployment instead of the in-repo one.


ChrisLovering (approved):
We should probably add limits to this at some point

PR #350, opened 2024-06-07
Add a preStop hook to site to close all connections

shtlrs:
This is to ensure that the connections opened from the pod that's about to be killed/replaced are closed beforehand.


shtlrs:
closing as this might mot even be needed

PR #354, opened 2024-06-07
Update site to run migrations in an init container

jb3:
In accordance with updates from python-discord/site#1338 this changes
the way migrations are run.

Previously, migrations would be run all from within the manage.py
execution process with the command being manually spawned using Django
internals.

After python-discord/site#1338 merges the Dockerfile will directly
invoke gunicorn and bypass manage.py to simplify the process and avoid
problems with shared database contexts.

Hence, we need to manually run migrations using an init container. With
testing there is no additional delay in doing this as spinning up an
init container is cheap and we don't cut over any traffic until the site
passes a healthcheck anyway.

### DIFF
--- a/kubernetes/namespaces/web/site/deployment.yaml
+++ b/kubernetes/namespaces/web/site/deployment.yaml
@@ -18,9 +18,9 @@ spec:
           image: ghcr.io/python-discord/site:latest
           imagePullPolicy: Always
           command:
-            - sh
-            - -c
-            - "poetry run python manage.py migrate"
+            - python
+            - manage.py
+            - migrate
           envFrom:
             - secretRef:
                 name: site-env
--- a/kubernetes/namespaces/web/site/deployment.yaml
+++ b/kubernetes/namespaces/web/site/deployment.yaml
@@ -13,6 +13,19 @@ spec:
       labels:
         app: site
     spec:
+      initContainers:
+        - name: migrations
+          image: ghcr.io/python-discord/site:latest
+          imagePullPolicy: Always
+          command:
+            - sh
+            - -c
+            - "poetry run python manage.py migrate"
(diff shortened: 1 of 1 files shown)


kubernetes/namespaces/web/site/deployment.yaml
```diff
@@ -13,6 +13,19 @@ spec:
       labels:
         app: site
     spec:
+      initContainers:
+        - name: migrations
+          image: ghcr.io/python-discord/site:latest
+          imagePullPolicy: Always
+          command:
+            - sh
```
MarkKoz, line 21: Why do we need a shell? Is it to resolve `poetry` on PATH?

jb3: Sanity check complete. I am insane.

jb3: Amended in 549a559

kubernetes/namespaces/web/site/deployment.yaml
```diff
@@ -13,6 +13,19 @@ spec:
       labels:
         app: site
     spec:
+      initContainers:
+        - name: migrations
+          image: ghcr.io/python-discord/site:latest
+          imagePullPolicy: Always
+          command:
+            - python
```
MarkKoz, line 21: Because we're setting the entrypoint to `poetry run`, we can omit it here, right?

MarkKoz (approved):

kubernetes/namespaces/web/site/deployment.yaml
jb3: Hmm, it is omitted here, this would be the equivalent of `[poetry run] python manage.py migrate` right?

PR #360, opened 2024-06-15
Add Kubernetes volume alerts

jb3:
It seems that Linode has added storage reporting info to the CSI driver
allowing us to pick up on the storage use of persistent volume claims
within the cluster.

This creates and deploys an alert that will report if any volume has
under 10% of space left. I have excluded Prometheus as our TSDB
retention settings mean that it will always stay just below it's volume
size by design.


jb3:
> 
> Doesn't that mean we can still include the alert, in case it breaks or
> something else fills up the disk? As far as I remember the TSDB retention size
> was a best effort: it works almost all of the time but sometimes you end up
> with higher disk usage.
> 

Yes, we just use over 90% of the disk for Prometheus, so right now it would be constantly alerting (hence my exclusion in the first place).

I don't know if it would be better to reduce the retention or exclude it.

We already have Grafana alerting if Prometheus becomes unhealthy.

PR #363, opened 2024-06-16
Add tooling documentation

jchristgit:


jb3 (approved):
Looks good, I would maybe explicitly add the `/merge` command to the documentation but it is in the policy file as mentioned, so no biggie.

jb3 (approved):

PR #369, opened 2024-06-17
Install tmux

jchristgit:


ChrisLovering (approved):

jb3:
I would like emacs-nox but I can add it myself or I can just edit over sudo SSH now.

PR #373, opened 2024-06-21
Deployment of a finger server

jb3:
This allows for external users to check on the status of the DevOps team using
the [finger](https://en.wikipedia.org/wiki/Finger_(protocol)) protocol.

Documentation is provided on establishing your plan as well as opting-out of the
finger system.

### DIFF
--- a/dns/zones/pydis.wtf.yaml
+++ b/dns/zones/pydis.wtf.yaml
@@ -2,10 +2,10 @@
 '':
   - octodns:
       cloudflare:
-        proxied: true
+        proxied: false
     ttl: 300
     type: A
-    value: 192.0.2.0 # Reserved placeholder IPv4 address
+    value: 89.58.25.151
   - octodns:
       cloudflare:
         auto-ttl: true


jb3:
After playing around, I cannot get this to work nicely unfortunately. It's just
a little too dated protocol and tooling wise and it just does not seem to work
on modern Debian.

jb3:
It was a nice idea and I'm not opposed to building a lightweight finger server
(similar to what `finger jb3@plan.cat` returns).

I chose `cfingerd` but this seemed to always hang when performing IDENT lookups
on port 113, which is not something that can be disabled (even allowing
anonymous lookups still triggers an ident lookup).

Even with the provided xinetd configuration from cfingerd the IP address that
cfingerd was picking up was just not the one that was actually requesting, I
think that the applications have diverged too much considering the age of
cfingerd.

Maybe another finger daemon would work better at this and return the information
we want, but I am also concious that finger daemons are notorious for being
exploit points or information exposure endpoints, so whatever we do I wanted to
ensure it was done right.

The client works completely fine, the server was the issue.

Maybe we can revisit this during the next DevOps meeting.

PR #403, opened 2024-07-08
Deploy rrdtool-based statistics

jchristgit:


ansible/roles/rrdstats/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+rrdstats_script_path: /opt/pydis/generate-rrdtool-stats.sh
+rrdstats_env_path: /etc/opt/pydis-rrdstats.env
+
+rrdstats_pg_username: rrdtool
```
shtlrs, line 5: all pg users have been put in the postgres role, might as well do the same here ?

ansible/roles/rrdstats/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+  ansible.builtin.template:
+    src: generate-rrdtool-stats.sh.j2
+    dest: "{{ rrdstats_script_path }}"
+    owner: root
+    group: root
+    mode: "0555"
+  tags:
+    - role::rrdstats
+
+- name: Configure PostgreSQL user
```
shtlrs, line 19: I think it makes sense to remove this and let the pg role handle it, since the pg role is already a dependency of this one.

ansible/roles/rrdstats/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    # -> https://github.com/ansible/ansible/issues/18955
+    # -> https://github.com/ansible-collections/community.general/issues/118
+    # -> https://github.com/ansible-collections/community.postgresql/issues/15
+    # Otherwise, we would need:
+    # - messages: id, channel_id, created_at
+    # - threads: id, archived
+    # - users: id, in_guild
+    # The `id` columns are almost purely needed for counting.
+    privs: SELECT
+    objs: messages,threads,users
```
shtlrs, line 60: same here

ansible/roles/rrdstats/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    password: "{{ rrdstats_pg_password }}"
+    state: present
+  tags:
+    - role::rrdstats
+
+- name: Configure PostgreSQL privileges to connect to database
+  become: true
+  become_user: postgres
+  community.postgresql.postgresql_privs:
+    database: metricity
```
shtlrs, line 35: this could also go under postgres_grants in pg role's vars, like we do with the rest.

shtlrs (changes requested):

shtlrs (approved):

PR #416, opened 2024-07-22
Add DNS entry for propaganda

jchristgit:
This commit adds a new DNS entry, used for storing Python
Discord-related propaganda, including, but not limited to, Soviet
propaganda, including, but not limited to, pictures of Joe Banks wearing
Ushankas and fan articles about Ivan Drago, as well as propaganda
displaying Joe Banks in various heroic settings.

### DIFF
--- a/dns/zones/pydis.wtf.zone/root.yaml
+++ b/dns/zones/pydis.wtf.zone/root.yaml
@@ -172,6 +172,14 @@ prometheus:
   type: CNAME
   value: linode-lb.box.pydis.wtf.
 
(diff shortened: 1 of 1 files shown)


jb3:
What are we planning on hosting here? This could be a little on-the-nose as far as our internal shitposting goes.

jb3:
I think we should consider calling it something other than `soviets` maybe, propaganda or morale or something probably is a bit more descriptive.

jb3 (approved):

PR #441, opened 2024-07-30
document linting workflow

jchristgit:
- Do not nest poetry in poetry
- Document local linting workflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,3 +33,13 @@ Infrastructure-related documentation ("the big picture"), can be found in [`docs
 
 Many folders have a `README.md` file within them, which have more detailed explanations on what
 that folder, and the files within, is used for.
+
+## Linting
+
+To lint everything with `pre-commit` locally, run the command
+
+```sh
+$ poetry run task lint
+```
+
+Any trivial errors (formatting, minor linting problems) will automatically be fixed.


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
     - id: ruff-format
       name: ruff formatting
       description: Run ruff formatting
-      entry: poetry run ruff format --force-exclude
+      entry: ruff format --force-exclude
```
ChrisLovering, line 28: The reason it was nested like this, and why we do this in other repos, is so that the pre-commit checks can still run even if the user hasn't acitvated their poetry shell.

EG if I open a terminal and try to commit a change to a python file I get this error.

```
13:34:46 ❯ git commit -m "foo"
check for merge conflicts................................................Passed
check toml...........................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
ruff linting.............................................................Failed
- hook id: ruff-lint
- exit code: 1

Executable `ruff` not found

ruff formatting..........................................................Failed
- hook id: ruff-format
- exit code: 1

Executable `ruff` not found
```
It's not easy for someone who doesn't already have this context to know that in order to commit changes I need to activate a poetry shell

jb3 (approved):

PR #442, opened 2024-07-30
Configure neomutt folder used by postfix

jchristgit:


ansible/roles/neomutt/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Configure neomutt folder to standard destination
+  copy:
+    content: |
+      # Ansible managed
+
+      # neomutt will user ~/Mail by default, whilst our postfix installation
+      # will deliver mail to /var/mail/$USER instead. neomutt will also figure that out,
+      # but not without prompting whether to create ~/Mail on every startup first.
+      set folder = /var/mail/$USER
```
jb3, line 17: I could be wrong, but what does the `folder` configuration option actually mean?

We don't use maildirs, we use a single mbox file, is that what this option is intended for?

jb3 (changes requested):

PR #450, opened 2024-08-03
Add a dns record for Jitsi

shtlrs:
We will be deploying jisti on lovelace soon, so this is needed in order to route traffic to our lovelace instance and use the service


jb3 (approved):

jb3:
Feel free to rebase and /merge at will

PR #455, opened 2024-08-08
Deliver mails as maildir

jchristgit:


jb3:
Nice. Do we need any sort of migration for existing mbox style directories?

jb3:
We tested this yesterday and right now cannot use this approach as Postfix will not create the maildirs.

Issue #460, opened 2024-08-13
E-Mail services


jchristgit:
Python Discord DevOps[^1] would like to provide a new service for our users, that
is, **e-mail based services**. Simply put, this would work as follows:

- User sends an e-mail to `$SERVICE@pydis.wtf` (or perhaps
  `$SERVICE@srv.pydis.wtf`)
- Postfix looks up `$SERVICE` in its tables
- Postfix dispatches to a fitting `master.cf` entry that spawns an external
  command taking the e-mail
- The external command processes the e-mail, and, if the service wants to,
  replies to the user.

**Suggestions for these commands**:

- Return a fortune cookie from the `fortune` command.
- Display current alerts from our Prometheus server, if any are present.
- Return an overview of PyDis infra issues that were updated recently.

For a start, we want to explore the concept, so I propose we simply add a
`fortune` service at first to test the feasability and maintainability of these
scripts. Scripts for these services should be maintained in Ansible.

We may or may not want to restrict access to these services for `@pydis.wtf`
users, or more specifically, DevOps team members. During configuration, care
must be taken to ensure that we do not allow spam through forged sender
addresses as a result of this.

[^1]: represented by me, and only me, with vague interest announced by other
    members of the team, excluding, this time, Bella, who has loudly complained
    about my constant misrepresentations of his actual intentions. I hereby
    subtly but with intent apologize for this. It was, in fact, not Bella who
    asked for the Sphinx documentation, the Kubernetes upgrade to alpha
    versions, or installation of Debian Experimental.

PR #461, opened 2024-08-13
Add service mail via. fortune address

jb3:
Partially addresses #460.

This is deployed and can be tested by sending an email (any subject and content)
to [`fortune@int.pydis.wtf`](mailto:fortune@int.pydis.wtf).

- Add a transport map to forward fortune@int.pydis.wtf to a pipe
- Add the pipe to master.cf
- Copy scripts and transport map in Postfix role

This also is implemented via a new `servicemailer` service account created in
LDAP which the mail scripts are run using, avoiding using the root or postfix
user (both are prohibited by Postfix).


ChrisLovering (approved):
Joe,

Just wanted to say a big thanks for this new service. I mean, who wouldn’t love the idea of sending an email just to receive a cryptic fortune in return? It’s like the perfect blend of nostalgia and modern inconvenience.

I'm really looking forward to typing out an email, hitting send, and then waiting to find out if I'm destined for "unexpected challenges" or "a surprise encounter". It's like a little slice of the 90s dial-up era, right in my inbox. And let’s be honest, my inbox was really missing that extra touch of mystery.

So, yeah, thanks for making email exciting again. Couldn't have imagined a more thrilling way to predict my future.

PR #462, opened 2024-08-14
Add docs on parsing mail in scripts

jb3:
Add some documentation on manual parsing of mail to avoid issues we saw with the
Fortune serving allowing for spoofing/unauthorized mail relaying.


jb3:
Agreed, maybe we can open an issue to track?

PR #463, opened 2024-08-15
Add documentation on how to ignore render[bot]

jchristgit:
In the latest episode of Joe Banks' continued war against our attention,
a new bot using the push-based user notification model has spawned on
the python-discord/infra repository. Attached my diplomatic response.

### DIFF
--- a/docs/docs/tooling/github-bots.md
+++ b/docs/docs/tooling/github-bots.md
@@ -40,7 +40,9 @@ receive:
 ``` sieve
 require ["envelope", "fileinto", "imap4flags"];
 
-if allof (header :is "X-GitHub-Sender" ["coveralls", "github-actions[bot]", "netlify[bot]"],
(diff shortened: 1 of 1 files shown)


jb3 (changes requested):
I'm not sure about blindly doing this.

Can we restructure this segment so instead it's a table of each potential ignore, what they do (i.e. render *just* builds doc previews on branches that change docs), and then the Sieve excerpt to ignore them?

Render is not noisy, it's one extra comment per PR and only if they change docs, so I think we should run this as an opt-out rather than blind ignore without reasoning? (I think blind ignoring is fine, but let's at least explain what you are ignoring so users can make an informed decision).

At the same time, I realise you are the only member of the DevOps team who uses/has the ability to use Sieve scripts so if you think it's fine then that's okay.

jb3:
Merged into main with modifications manually.

PR #471, opened 2024-08-18
Add a role to install Jitsi.

shtlrs:
Since the beginning of time, we have relied on dennis services to host our very productive devops meetings, thanks Dennis.

However, a new dawn shines its light on the Pydis Devops team, a dawn where we have our own service to host our invaluable meetings.

This PR adds the role that configures and install jitsi to run on `jitsi.pydis.wtf`.

### DIFF
--- a/ansible/roles/jitsi/meta/main.yml
+++ b/ansible/roles/jitsi/meta/main.yml
@@ -1,0 +1,4 @@
+---
+dependencies:
+  - nftables
+  - nginx
--- a/ansible/roles/jitsi/tasks/main.yml
+++ b/ansible/roles/jitsi/tasks/main.yml
@@ -1,0 +1,101 @@
+---
+- name: Download and add Prosody Debian packages key
+  ansible.builtin.get_url:
+    url: https://prosody.im/files/prosody-debian-packages.key
+    dest: /etc/apt/keyrings/prosody-debian-packages.key
+    mode: '0644'
+  tags:
+    - role::jitsi
+
+- name: Add Prosody repository to sources list
+  ansible.builtin.apt_repository:
+    repo: "deb [signed-by=/etc/apt/keyrings/prosody-debian-packages.key] http://packages.prosody.im/debian {{ ansible_distribution_release }} main"
+    filename: prosody-debian-packages
+  tags:
+    - role::jitsi
+
+- name: Install lua5.2
+  ansible.builtin.apt:
+    name: lua5.2
+    state: present
+  tags:
+    - role::jitsi
+
+
+- name: Fetch Jitsi GPG key
+  ansible.builtin.get_url:
+    url: https://download.jitsi.org/jitsi-key.gpg.key
+    dest: /tmp/jitsi-key.gpg.key
+    mode: "u=rw,g=r,o=r"
+  tags:
+    - role::jitsi
+
+- name: Convert GPG key to keyring format
+  ansible.builtin.command:
+    cmd: gpg --dearmor -o /etc/apt/keyrings/jitsi-keyring.gpg /tmp/jitsi-key.gpg.key
+    creates: /etc/apt/keyrings/jitsi-keyring.gpg
+  tags:
+    - role::jitsi
+
+- name: Clean up temporary GPG key file
+  ansible.builtin.file:
+    path: /tmp/jitsi-key.gpg.key
+    state: absent
+  tags:
+    - role::jitsi
+
+- name: Add Jitsi repository to sources list
+  ansible.builtin.apt_repository:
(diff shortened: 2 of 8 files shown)


ansible/roles/jitsi/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+- name: Add Jitsi repository to sources list
+  ansible.builtin.apt_repository:
+    repo: "deb [signed-by=/usr/share/keyrings/jitsi-keyring.gpg] https://download.jitsi.org stable/"
+    filename: jitsi-stable
+  become: true
+  tags:
+    - role::jitsi
+
+- name: Preconfigure debconf settings for Jitsi
+  debconf:
```
jb3, line 62: Huh, this is awesome.

ansible/roles/jitsi/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+    dest: /tmp/jitsi-key.gpg.key
+    mode: "u=rw,g=r,o=r"
+  become: true
+  tags:
+    - role::jitsi
+
+- name: Convert GPG key to keyring format
+  ansible.builtin.command:
+    cmd: gpg --dearmor -o /usr/share/keyrings/jitsi-keyring.gpg /tmp/jitsi-key.gpg.key
+    creates: /usr/share/keyrings/jitsi-keyring.gpg
```
jb3, line 40: For Grafana, we write this keyring to `/etc/apt/keyrings/grafana.gpg`, can we put the Jitsi one there also for consistency?

ansible/roles/jitsi/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
+---
```
jb3, line 1: Is there a reason all the tasks in this role use `become: true`? It shouldn't be necessary as we specify it in our ansible.cfg

ansible/roles/jitsi/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+    state: present
+  become: true
+  tags:
+    - role::jitsi
+
+- name: Activate the jitsi server block
+  ansible.builtin.file:
+    src: /etc/nginx/sites-available/jitsi.pydis.wtf.conf
+    path: /etc/nginx/sites-enabled/jitsi.pydis.wtf.conf
+    state: link
```
jb3, line 85: You should use the `reload the nginx service` handler here, it also needs a tag.

ansible/roles/jitsi/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+# https://community.jitsi.org/t/ssl-handshake-error-sslv3-alert-certificate-unknown/41245
+
+- name: Disable Video Bridge certificate verification
+  lineinfile:
+    dest: /etc/jitsi/videobridge/sip-communicator.properties
+    line: org.jitsi.videobridge.xmpp.user.shard.DISABLE_CERTIFICATE_VERIFICATION=true
+    state: present
+    create: false
+    owner: jvb
+    group: jitsi
```
jb3, line 99: Needs a tag.

ansible/roles/jitsi/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+dependencies:
+  - nftables
```
jb3, line 3: I don't know that this actually needs to be a dependency, it will just update when we run the nftables role and that's fine.

NGINX should be as you'll need to use the handlers (see other comment)

jb3 (changes requested):

ansible/roles/jitsi/meta/main.yml
shtlrs: I just thought that we'd run the role only when we've had all the preconfiguration necessary, i can remove it if you'd like

ansible/roles/jitsi/tasks/main.yml
shtlrs: 0491979

ansible/roles/jitsi/tasks/main.yml
shtlrs: 8c6538a

ansible/roles/jitsi/tasks/main.yml
shtlrs: Neh i think not, removed here a3eda10

ansible/roles/jitsi/tasks/main.yml
shtlrs: Oui bien sur, 57b2655

jb3 (approved):
ROCK SOLID!

PR #482, opened 2024-08-22
simplify dovecot config

jchristgit:
- Automatically configure default IMAP folders
- Fix "an actual problem"
- Add archive & automatically create mailboxes
- Simplify dovecot configuration templating logic


jb3:
Cherry picked in

PR #483, opened 2024-08-23
Run saslauthd under the "sasl" user & group

### LINKED ISSUE #478: Restrict saslauthd user
`saslauthd` currently runs as root. This is bad.

`man saslauthd` documents that:

>    saslauthd should be started from the system boot scripts when going to multi-user mode. When run‐
>    ning against a protected authentication database (e.g. the shadow mechanism), it must be run as
>    the superuser. Otherwise it is recommended to run daemon unprivileged as saslauth:saslauth, re‐
>    quiring the runtime directory to have root:saslauthd owner. You can do so by following these steps
>    in machines using systemd(1) :

>    1.   create directory /etc/systemd/system/saslauthd.service.d/
>    2.   create file /etc/systemd/system/saslauthd.service.d/user.conf with content
>
>         [Service]
>         User=saslauth
>         Group=saslauth
>
>    3.   Reload systemd service file: run “systemctl daemon-reload”

After discussion with Joe, we do not use `saslauthd` for shadow auth.

We should therefore implement this to improve system security.

MarkKoz:
Resolves #478

To improve security, run it as a new "sasl" user rather than as root. The existing "sasl" group is re-used.

Note that Ansible does not provide an agnostic way to configure the user and group for a service. This change couples the configuration to systemd.

### Testing
It's difficult for me to test this because I'm on Windows. If someone else can test this, I'd appreciate it.


ansible/roles/sasl/tasks/main.yml
```diff
@@ -64,7 +84,7 @@
     - role::sasl
 
 - name: Enable and start SASL authentication daemon
-  service:
+  systemd_service:
```
shtlrs, line 87: Couldn't this be acheived by running the service as a paricular user?

meaning

```py

become: true
become_user: sasl
```

However, this might imply that we need to add that user to sudoers, but only give it the righ to run that service only.


Also, can't we just modify the existing unit file ? you can use the `lineinfile` or `blockinfile` to replace the default one, and then just run the sevice.

shtlrs: I also don't think that coupling to systemd is an issue, i don't see us using something else than systemd to run our services.

We will see what the others have to say about it.

ansible/roles/sasl/tasks/main.yml
```diff
@@ -64,7 +84,7 @@
     - role::sasl
 
 - name: Enable and start SASL authentication daemon
-  service:
```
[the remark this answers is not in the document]
MarkKoz: Because we're configuring the user and group in a way that's specific to systemd. If this service was launched with something other than systemd, then that config would silently do nothing.

ansible/roles/sasl/tasks/main.yml
MarkKoz: No, I believe that would change the user of systemd itself. @jchristgit left an excellent answer.

I don't have the contents of the existing unit file, and that approach seems more heavy-handed. Would we have to dupliucate the contents of that file and then append the user/group config to it? Adding `user.conf` seems neater.

shtlrs: Ah nice i didn't know that!

Alright I see, thank boss

ansible/roles/sasl/tasks/main.yml
```diff
@@ -9,6 +9,26 @@
...
+  template:
+    src: user.conf
+    dest: /etc/systemd/system/saslauthd.service.d/user.conf
+    mode: "0640"
+    owner: root
+    group: root
+  tags:
+    - role::sasl
+  notify:
+    - Reload SASL
```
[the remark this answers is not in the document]
MarkKoz: You wrote "restart" SASL but the notify is still for a "reload". Should we change it to restart SASL, or did you mean to write reload?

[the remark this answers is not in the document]
MarkKoz: Addressed by the latest push c658a51

PR #484, opened 2024-08-23
Clean up SASL logic from postfix

jchristgit:
Per Joe's comment here [1].

[1]: https://github.com/python-discord/infra/issues/478#issuecomment-2306107672


jb3:
Do you want to retarget this onto my branch? Given that the dovecot PR is what removes this functionality.

PR #486, opened 2024-08-23
Add spamassassin

jb3:
Adds spamassassin to inbound mail passing through the PyDis mailserver.

This adds a variety of headers which can be used by end user email clients or by
Sieve to filter mail according to the determined importance.


jb3:
FIRE!

PR #493, opened 2024-08-25
Bind NGINX hosts on IPv6 as well

jchristgit:
Previously only Jitsi would answer when requesting on IPv6.


jb3 (approved):

jb3:
Merged manually in b9bceb4. Thanks!

PR #495, opened 2024-08-25
Unify alertmanager naming

### LINKED ISSUE #451: Unify AlertManager alert naming
Whilst we use the same alerting policies for both Kubernetes and Netcup, we haven't yet unified the naming.

Kubernetes is styled as `ContainerOOMEvents` whilst Netcup has something like `node/systemd-unit-failed`.

We should rename the Kubernetes alerts to use the latters styling. We can prefix the alerts with the group they are in (e.g. `coredns`, `cert-manager`) followed by the previous alert name but as a slug (e.g. `KubernetesPodCrashLooping` to `kubernetes/pod-crash-looping`)

jchristgit:
Closes #451.

### DIFF
--- a/kubernetes/namespaces/monitoring/alerts/alerts.d/alertmanager.yaml
+++ b/kubernetes/namespaces/monitoring/alerts/alerts.d/alertmanager.yaml
@@ -2,7 +2,7 @@ groups:
 - name: alertmanager
   rules:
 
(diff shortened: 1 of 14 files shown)


jb3 (approved):
ROCKSTAR!

jb3:
Merged manually in ba5208b. Thanks!

PR #502, opened 2024-08-26
Install prometheus-postfix-exporter

jchristgit:
As a data-obsessed administrator I want to have more data such that I
can widen my sense of power.

This also installs rsyslog, because prometheus-postfix-exporter doesn't
work with journald's binary log format.

### DIFF
--- a/ansible/host_vars/lovelace/prometheus.yml
+++ b/ansible/host_vars/lovelace/prometheus.yml
@@ -55,6 +55,13 @@ prometheus_configuration: |
           - {{ hostvars[host]['ansible_wg0']['ipv4']['address'] }}:9187
           {%- endfor %}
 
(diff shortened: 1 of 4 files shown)


jb3 (approved):
ROCK SOLID!

jb3 (changes requested):
Actually, the alerting here won't work.

Prometheus handles alerts for it's own instance, and the instance scraping here is the one on lovelace.

Alerts need to be defined in the `prometheus_rules` section of `ansible/host_vars/lovelace/prometheus.yml`

jb3 (approved):

jb3:
Merged in efb9fa9194ff5f70fe027b945787cbc34293db05. Thanks!

PR #503, opened 2024-08-26
Stop installing libsasl2-modules-ldap

jchristgit:
This caused problems in postfix's logfile, our most holy daemon. We must
not disturb it, and as an aside, everything works fine.

### DIFF
--- a/ansible/roles/jitsi/tasks/ldap_auth.yml
+++ b/ansible/roles/jitsi/tasks/ldap_auth.yml
@@ -47,7 +47,6 @@
     state: present
   loop:
     - sasl2-bin  # Necessary for Cyrus' saslauthd
-    - libsasl2-modules-ldap  # Necessary for Cyrus' saslauthdp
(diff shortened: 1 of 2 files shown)


jb3 (approved):

jb3:
Merged in a31be18.

shtlrs:
What are the consequences of removing this for LDAP auth when using jitsi ?

If i uninstall the package, and delete jitsi then redeploy, would it work still ?

shtlrs:
> What are the consequences of removing this for LDAP auth when using jitsi ?
> I don't know, because the package is already uninstalled, unless someone reinstalled it recently. Joe removed it recently and all it did was prevent some error from spamming The Holy Postfix Logfile.

Hum that's strange , because the official JITSI LDAP docs say that it needs to be installed in order for LDAP Auth to work.

shtlrs:
It seems to work, just tested it. 🤷

PR #504, opened 2024-08-26
ManageSieve for Everyone

jchristgit:
A spectre is haunting Python Discord -- the spectre of proprietary mail
filtering programs. All the Powers of Big Mail have entered into a holy
alliance to exorcise this spectre: Outlook and GMail, Yahoo and Zoho,
AOL adn iCloud.

Where is the open protocol in opposition that has not been decried as
unprofessional, free and open source by its opponents in power? Where is
the Opposition that has not hurled back the branding reproach of
Internet Standards, against the more advanced opposition protocols, as
well as against its reactionary adversaries?

Two things result from this fact.

I.  Sieve is already standardized as a protocol via the IETF.

II. It is high time that Python Discord users should openly, in the face
of the TLS-encrypted internet, filter their notifications, their love
letters, their mailing lists, and meet this nursery tale of the Spectre
of ManageSieve with a Manifesto of Python Discord Devops itself.

To this end, ManageSieve users of various nationalities have assembled
on jitsi.pydis.wtf, and sketched the following commit, to be published
on the lovelace Python Discord e-mail service.

### DIFF
--- a/ansible/roles/dovecot/templates/dovecot.conf.j2
+++ b/ansible/roles/dovecot/templates/dovecot.conf.j2
@@ -87,12 +87,6 @@ dict {
   #quota = mysql:/etc/dovecot/dovecot-dict-sql.conf.ext
 }
 
-# PYDIS CONFIG START
-
-protocols = imap lmtp
-
-# PYDIS CONFIG END
-
 # Most of the actual configuration gets included below. The filenames are
 # first sorted by their ASCII value and parsed in that order. The 00-prefixes
 # in filenames are intended to make it easier to understand the ordering.
--- a/docs/docs/services/email/mail-clients.md
+++ b/docs/docs/services/email/mail-clients.md
@@ -60,15 +60,14 @@ are not provided by the PyDis mailserver please let us know in `#dev-oops`.
 
 We support server-side email filtering with [Pigeonhole
 Sieve](https://doc.dovecot.org/configuration_manual/sieve/pigeonhole_sieve_interpreter/).
+Sieve scripts are managed via [the ManageSieve
+protocol](https://datatracker.ietf.org/doc/html/rfc5804). Your e-mail client
+should have built-in functionality for writing and editing these scripts. For a
+listing of clients, see [the official Sieve website](http://sieve.info/clients).
 
-!!! example "DevOps Only"
-
-    For now, since this feature requires access to a home directory on the host, it is
-    only accessable to members of the DevOps team. If you are not a member of the DevOps
-    team and wish to configure sieve, please let us know in the #dev-oops channel.
-
-    We may in future investigate further usage of `dovecot-managesieved` to allow for
-    remote management of the Sieve filters.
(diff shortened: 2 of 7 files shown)


docs/docs/services/email/mail-clients.md
```diff
@@ -60,15 +60,14 @@ are not provided by the PyDis mailserver please let us know in `#dev-oops`.
...
-    For now, since this feature requires access to a home directory on the host, it is
-    only accessable to members of the DevOps team. If you are not a member of the DevOps
-    team and wish to configure sieve, please let us know in the #dev-oops channel.
-
-    We may in future investigate further usage of `dovecot-managesieved` to allow for
-    remote management of the Sieve filters.
+Sieve scripts are managed via [the ManageSieve
+protocol](https://datatracker.ietf.org/doc/html/rfc5804). Your e-mail client
+should have built-in functionality for writing and editing these scripts. For a
+listing of clients, see [the official Sieve website](sieve.info/clients.).
```
jb3, line 66: This link renders wrong, also, sieve.info only works over HTTP:
```suggestion
listing of clients, see [the official Sieve website](http://sieve.info/clients).
```

ansible/roles/dovecot/templates/dovecot.conf.j2
```diff
@@ -87,12 +87,6 @@ dict {
   #quota = mysql:/etc/dovecot/dovecot-dict-sql.conf.ext
 }
 
-# PYDIS CONFIG START
-
-protocols = imap lmtp
-
-# PYDIS CONFIG END
-
```
jb3, line 95: I'm sure there is a reason, but why is this diff removed?

jb3 (changes requested):

MarkKoz (commented):
Have I understood correctly that our mail server already supports sieve, but we were merely blocking it in our firewall?

We're using Dovecot, so do we have the [sieve plugin](https://doc.dovecot.org/2.3/configuration_manual/sieve/#sieve) already installed and configured?

jb3:
> Have I understood correctly that our mail server already supports sieve, but we were merely blocking it in our firewall?

Correct, it came in the Dovecot PR but we didn't enable it.

> We're using Dovecot, so do we have the [sieve plugin](https://doc.dovecot.org/2.3/configuration_manual/sieve/#sieve) already installed and configured?

Yes, Sieve is installed and configured. By default, it's looking for config in the `~/sieve` file. I don't know if this PR needs us to specify the additional parameters in the Sieve user attribute for ManageSieve support (i.e. `active`)

jb3 (changes requested):
Works great.

Can we recommend https://github.com/thsmi/sieve as a client as well as sieve-connect?

I've tried this myself and not only does it work great but it also has a nice graphical overview of rule processing which might be easier for non-technical users (as well as of course allowing for manual sieve editing).

jb3 (approved):

PR #505, opened 2024-08-26
Feed Dovecot maildir stats into Prometheus

jchristgit:
Already deployed on lovelace. I was going to leave out script deployment
and just write it inline, but YAML folding of long lines (and
unreadability of 200 column-wide AWk scripts, to be honest) made it a
bit bad.

The e-mail for DevOps cron failure reports is updated to include `+cron`
to allow for client-side filtering, if necessary.

To test: `ssh -L localhost:9090:localhost:9090 lovelace.box.pydis.wtf`,
then check out the `dovecot_` variables in the UI.

We might want to further check out Dovecot's built-in statistics
support, see https://doc.dovecot.org/2.3/configuration_manual/stats/.

### DIFF
--- a/ansible/roles/dovecot-monitoring/tasks/main.yml
+++ b/ansible/roles/dovecot-monitoring/tasks/main.yml
@@ -1,0 +1,52 @@
+---
+- name: Create dovecot monitoring directory
+  ansible.builtin.file:
+    path: "{{ dovecot_monitoring_scripts_directory }}"
+    state: directory
+    owner: root
+    group: root
+    mode: "0755"
+  tags:
+    - role::dovecot-monitoring
+
+- name: Create dovecot monitoring scripts
+  ansible.builtin.template:
+    src: "{{ item }}.j2"
+    dest: "{{ dovecot_monitoring_scripts_directory }}/{{ item }}"
+    owner: root
+    group: root
+    mode: "0544"
+  tags:
+    - role::dovecot-monitoring
+  loop:
+    - maildir-mails.sh
+    - maildir-sizes.sh
+
+- name: Create Maildir size monitoring cronjobs
+  ansible.builtin.cron:
+    name: "{{ item.name }}"
+    minute: "*/20"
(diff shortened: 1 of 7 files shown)


ansible/roles/dovecot-monitoring/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+  loop:
+    - name: Dovecot maildir size Prometheus exporter
+      job: "{{ dovecot_monitoring_scripts_directory }}/maildir-sizes.sh"
+    - name: Dovecot maildir mail count Prometheus exporter
+      job: "{{ dovecot_monitoring_scripts_directory }}/maildir-mails.sh"
+
+- name: Report to DevOps when Maildir size exporter fails
+  ansible.builtin.cron:
+    name: MAILTO
+    env: yes
```
jb3, line 47: Ansible Lint is complaining about this line:
```suggestion
    env: true
```

jb3 (changes requested):

jb3 (changes requested):
Another minor thing with this: the `dovecot_maildir_size_bytes` metric is not actually exposing bytes, it's exposing kilobytes, on account of the fact that `du`, by default, outputs units of 1024 bytes:

```
joe@lovelace:~$ sudo du -s /var/vmail/joe
5260    /var/vmail/joe
joe@lovelace:~$ sudo du -sh /var/vmail/joe
5.2M    /var/vmail/joe
```

jb3:
Grafana dash for email: https://grafana.pydis.wtf/d/edw0wuqz1yccgf/email?orgId=2

ansible/roles/dovecot-monitoring/tasks/main.yml
MarkKoz: YAML is brilliant

jb3 (approved):

jb3:
Merged into main. Thanks!

PR #516, opened 2024-08-29
Implement dovecot mail plugin, configure limits

jchristgit:
This also moves custom plugins into the `roles/munin/templates/plugins`
directory, which should hopefully be easier to maintain than the
existing inline dictionary. The only issue is that now it is a bit
harder to deal with the filepaths.

This change has already been deployed.


jb3:
Merged into main. Thanks!

PR #517, opened 2024-08-29
Update git mirror once a day

jchristgit:
We do not commit code nearly enough to warrant us checking for updates
on all repositories every day. I would even go as far as saying that
some repositories are updated rarely enough that they do not need to be
checked every day either.

### DIFF
--- a/ansible/roles/git-mirrors/tasks/main.yml
+++ b/ansible/roles/git-mirrors/tasks/main.yml
@@ -146,8 +146,9 @@
 - name: Add cronjob for mirror updating
   cron:
     name: "Update the git mirrors published by cgit (git-mirrors role)"
-    # Every 5 minutes
-    minute: "*/5"
+    # Every day at 4:27 AM
+    hour: "4"
+    minute: "27"
     job: "chronic {{ git_mirrors_base_dir }}/update-mirrors.sh"
     user: git-mirrors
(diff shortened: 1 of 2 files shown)


jb3:
Checking is incredibly cheap, it's a single HTTP call and no CPU cycles really get wasted.

It's been nice seeing things show up on cgit, it's not really a waste of resources.

If we want to make the script more advanced and poll different repositories at different intervals, that's fine. Likewise if we want a frequency this slow we should add a way to trigger an update with a webhooks, I don't really see an issue with a dozen or so HTTP requests every few minutes.

jb3:
To add, just how miniscule this operation is:
```
joe@lovelace:/srv/git-mirrors/mirrored/jc/poetry-restrict-plugin$ sudo -u git-mirrors time git fetch -q --prune
0.04user 0.01system 0:00.08elapsed 67%CPU (0avgtext+0avgdata 17296maxresident)k
```

jb3:
Can we re-specify this issue/PR to support dynamic mirror updates on a GitHub webhooks?

Frequency at once per day removes most of the gun, novelty and personally, the use.

I also don't buy the upstream respect, many similar projects pull at the exact same interval, https://github.com/pyouroboros/ouroboros comes to mind (despite it's deprecation).

PR #521, opened 2024-08-31
Document alertmanager reloads

jchristgit:


jb3 (changes requested):
Implying the Makefile will immediately reload the AlertManagers is a bit of a red herring.

Due to the delay in ConfigMap rollout when a ConfigMap is applied it takes 2ish minutes for the file mounted to the pod to change.

We need to instead configure AlertManager like we do Prometheus using inotify-base and then reload once we recognise the file changing. This right now won't actually reload the new config into the AlertManager (or it will if run at a staggered time after applying the new config, but this isn't obvious to the end user).

We can use the Prometheus setup for inspiration here, it's almost the exact same configuration for the sidecar container.

PR #522, opened 2024-08-31
Raise time threshold for 4xx alerts

jchristgit:
At present we get plenty of unactionable, flapping alarms. So far, they
have shown us nothing of value. Raise the time consecutive errors need
to be seen before we alert.


jb3 (approved):

jb3:
Merged in 276645b.

PR #524, opened 2024-09-01
Swap saslauthd to mount outside of postfix

### LINKED ISSUE #474: Swap to a shared mount when Postfix is moved to Dovecot
We currently have a hack in place to have both jitsi and postfix work with the same saslauth config

https://github.com/python-discord/infra/blob/aea5fdbe2f9815b3ced9fead7f19605268f319ee/ansible/roles/jitsi/tasks/ldap_auth.yml#L63-L70

We should swap to a shared mount when Postfix is moved to Dovecot.

jchristgit:
Closes #474.

### DIFF
--- a/ansible/roles/sasl/tasks/main.yml
+++ b/ansible/roles/sasl/tasks/main.yml
@@ -51,6 +51,29 @@
   notify:
     - Restart SASL
 
+# The sysvinit script checks for a dpkg-statoverride when first starting. If
+# one exists, it uses it, if none exists, ti creates the directory with
+# ownership u=root g=sasl m=710. With these two tasks, we check whether there's
+# a default override already specified (presumably from package installation)
+# and only set it if it's distinct, ensuring idempotency.
+- name: Check current overrides for runtime directory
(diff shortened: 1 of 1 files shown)


ansible/roles/sasl/tasks/main.yml
```diff
@@ -45,12 +45,35 @@
...
+# a default override already specified (presumably from package installation)
+# and only set it if it's distinct, ensuring idempotency.
+- name: Check current overrides for runtime directory
+  command: dpkg-statoverride --list "{{ sasl_mux_path }}"
+  changed_when: false
+  register: sasl_statoverride_list
+  tags:
+    - role::sasl
+
+- name: Ask dpkg to create runtime directory as sasl user  # noqa: no-changed-when
```
jb3, line 66: Would it not make sense to use `changed_when: true` here?

jb3:
Potentially, I'm just going by the advised remediations in the ansible-lint docs so we wouldn't have to noqa, `changed_when: true` is what we do elsewhere.

https://ansible.readthedocs.io/projects/lint/rules/no-changed-when/

PR #528, opened 2024-09-01
Add all accepted mail domains to SRS exclude configuration

jb3:
We don't want to rewrite the envelopes of mail that is from a valid
domain of our mailserver (e.g. pydis.com or int.pydis.wtf), but by
default PostSRSD will rewrite anything that is not the `mydomain`
configuration variable of Postfix (which is just set to `pydis.wtf` for
us).

This change updates the environment defaults for PostSRSD to ensure that
we don't change any envelopes that don't need to be changed, as well as
ensuring that to end mailservers the DKIM and SPF checks are made
against the actual domain (e.g. int.pydis.wtf) instead of the rewritten
envelope.

### DIFF
--- a/ansible/roles/postfix/handlers/main.yml
+++ b/ansible/roles/postfix/handlers/main.yml
@@ -20,3 +20,8 @@
 - name: Regenerate sender access table
   command: postmap /etc/postfix/sender_access
   changed_when: true
+
+- name: Restart postsrsd
+  service:
+    name: postsrsd
+    state: restarted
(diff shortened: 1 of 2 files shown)


jb3:
> We should probably also add `role::postfix` to the handlers

Hmmm, we don't for other ones.

I believe the reason is that if we want to use a role as a dependency then
adding tags onto it stops the handlers running properely if the handlers
that have been notified are from a dependency that's not been tagged?

jb3:
Hmm, well they run right now without tags, and of course run with tags, and they'll only ever run if called from a task which does have a tag.

PR #531, opened 2024-09-02
Refactor "Managed by Ansible" notices in templates

jb3:
As of now, we have no coherence between "Managed by Ansible" and have a mix of
this string and "Ansible Managed".

We also have no indication on the remote filesystem to which role a given
templated file belongs or how to locate it once it's deployed.

I propose that we change to using the built-in `{{ ansible_managed }}` tag on
any template that we wish to signify is managed by our Ansible setup. On the
remote systems, this renders something like this:

``` sh
joe@lovelace:~$ cat /etc/nginx/conf.d/charset.conf
# Managed by Ansible. Belongs to nginx role.
charset utf-8;
```

I haven't finalised the format or which variables to include (though for obvious
reasons they should be static variables), hence why this is a draft.

I also haven't yet gone through to update templates with this change and re-run
all roles.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -5,6 +5,8 @@ vault_password_file = vault_passwords
 collections_path = .ansible/galaxy_collections/
 roles_path = .ansible/galaxy_roles/:roles/
 
+ansible_managed = Managed by Ansible (do not edit). Role: {{{{ role_name }}}}, Template: {{{{ template_path | basename }}}}
+
 [privilege_escalation]
 become = yes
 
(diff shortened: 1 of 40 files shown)


shtlrs:
This is a good idea.

However, how do you expect to handle 2 block in file replacements of the same file when you don't have a template? Or is this just for the files for which we have a template?

jb3:
> This is a good idea.
> 
> However, how do you expect to handle 2 block in file replacements of the same file when you don't have a template? Or is this just for the files for which we have a template?

I think for now we are fine to just handle templates, the blockinfiles are normally simple enough to understand and have constant marks.

PR #535, opened 2024-09-03
Clean up mail script after feeding it to at

jchristgit:


jb3 (approved):

jb3:
Applied manually to main, thanks!

PR #546, opened 2024-09-13
Install & configure pg_repack

ChrisLovering:


ChrisLovering:
> We can add this but I wonder if we need it? Do we have a lot of bloat?

I know with metricity specifcally I was repacking weekly on linode due to bloat, and it significantly improved performance. That might have been due to autovac having issues running in the container though.

ChrisLovering:
> Should we monitor bloat somehow? I'm happy to merge this either way but I think monitoring bloat might be useful.

Yeah that could be cool, not sure whether through rrdtool or prom, but having some chart somewhere would be good.

PR #550, opened 2024-09-19
CI user

jb3:
- Add a new CI user role
- Add restricted keys to this user to allow for rsyncing to the `/var/www` folder
- Update some nginx configurations to serve the new landing page for pydis.wtf


jb3:
> Shouldn't this be restricted to /var/www/pydis.wtf?

My thinking was we can reuse this user & key for other sites we want to deploy into NGINX instead of a webscale platform.

I suppose we could use a different key for each service.

PR #605, opened 2025-09-28
Add the codejam winners to the k8s cluster

ChrisLovering:
These will be deleted once Johannes next takes his meds


fisher60:
@ChrisLovering this is a reminder to take these services down when you are home.

GDWR:
@ChrisLovering this is a reminder to take these services down when you are home.

PR #657, opened 2026-05-12
Be kind to GitHub

jchristgit:
The Python Discord Git Mirror Infrastructure Mirror Cronjob is causing
repeated GitHub outages. Be kinder.


jb3:
No longer needed & has not been an issue recently. If it continues to be we should add a backoff before alerting.

PR #658, opened 2026-05-15
Tailscale Operator & Ansible Tailscale deployment

jb3:
This PR introduces Tailscale to both Kubernetes & Ansible hosts.

On Kubernetes, the Tailscale Operator is used to allow routing traffic in/out of the cluster over Tailscale to other nodes connected to the tailnet. This allows us to securely forward traffic to an external service from any Kubernetes node without having to compromise on security.

Likewise in Ansible we now provision Tailscale on every host, optionally configuring service forwarding (where a service is allocated it's own FQDN within the tailnet) for traffic to be routed to. We can use this to secure inter-node communication on Ansible hosts or bring traffic from Kubernetes safely into Ansible with minimal overhead.


ansible/roles/postgres/templates/postgresql.conf.j2
```diff
@@ -2,7 +2,7 @@ data_directory = '/var/lib/postgresql/{{ postgres_version }}/main'
 hba_file = '/etc/postgresql/{{ postgres_version }}/main/pg_hba.conf'
 ident_file = '/etc/postgresql/{{ postgres_version }}/main/pg_ident.conf'
 external_pid_file = '/var/run/postgresql/{{ postgres_version }}-main.pid'
-listen_addresses = '89.58.26.118,localhost'
+listen_addresses = '89.58.26.118,lovelace.opossum-python.ts.net,localhost'
```
jchristgit, line 5: I assume this means PostgreSQL won't start if Tailscale is not running, doesn't it?
Should we perhaps change this to `*` while we're at it, then? Since we already listen on our public IPv4 and rely on nftables there, we might as well, no?

ansible/roles/tailscale/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+  when: ansible_facts["distribution"] == "Debian"
+  tags:
+    - role::tailscale
+
+- name: Update APT cache (Debian)
+  ansible.builtin.apt:
+    update_cache: yes
+  when: ansible_facts["distribution"] == "Debian"
+  tags:
+    - role::tailscale
```
jchristgit, line 45: Suggestion 1: use a handler and then `- meta: flush_handlers` for this
Suggestion 2: Register the task above, then add `add_tailscale_apt_repository is changed` to `when` here

ansible/roles/tailscale/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+    dest: /etc/apt/sources.list.d/tailscale.list
+    mode: "0644"
+  when: ansible_facts["distribution"] == "Debian"
+  tags:
+    - role::tailscale
+
+- name: Update APT cache (Debian)
+  ansible.builtin.apt:
+    update_cache: yes
+  when: ansible_facts["distribution"] == "Debian"
```
jchristgit, line 43: We could group all of these together in a `block` for a single condition per distribution, but I am indifferent about it.

ansible/roles/tailscale/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+    tailscale_hosted_services: "{{ tailscale_services_status.stdout | from_json | json_query('services') }}"
+  tags:
+    - role::tailscale
+
+- name: Set tailscale_hosted_services to empty list if not defined
+  ansible.builtin.set_fact:
+    tailscale_hosted_services: []
+  when: not tailscale_hosted_services
+  tags:
+    - role::tailscale
```
jchristgit, line 109: Is there a way we could get the task above to return this? When would this condition hit?

ansible/roles/tailscale/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+
+- name: Ensure Tailscale services are configured
+  ansible.builtin.command: |-
```
jchristgit, line 112: The linter is right here, we should add `changed_when` depending on the item standard output.

ansible/group_vars/all/nftables.yml
```diff
@@ -51,6 +51,9 @@ nftables_configuration: |
       # Allow loopback
       iif lo accept
 
+      # Allow all traffic from/to tailscale0
+      iifname tailscale0 accept
```
jchristgit, line 55: This comment is incorrect, this is the input chain. It also restates what the code speaks.

```suggestion
      # Allow all traffic from Tailscale
      iifname tailscale0 accept
```

kubernetes/namespaces/tailscale/README.md
```diff
@@ -0,0 +1,20 @@
...
+2. Update the Helm repo `helm repo update`
+3. Install the tailscale operator, replacing OAuth credentials as necessary (from the Trust credentials section of Tailscale admin console):
+    ```bash
+    helm upgrade \
+    --install \
+    tailscale-operator \
+    tailscale/tailscale-operator \
+    --namespace=tailscale \
+    --create-namespace \
+    --set-string oauth.clientId="<OAauth client ID>" \
```
jchristgit, line 17:
```suggestion
    --set-string oauth.clientId="<OAuth client ID>" \
```

jchristgit (changes requested):
Looks good for the most part.

PR #660, opened 2026-05-18
Move tls secrets to nginx-gateway namespace

ChrisLovering:
These secrets are only consumed by nginx now, so no need to reflect them to other namespaces


jchristgit (approved):
I assume we also need to manually remove the existing ones?

PR #663, opened 2026-05-24
Add Vector & update Kubewatch configuration

### LINKED ISSUE #347: Retire the #dev-ops-notifications channel
Our `#dev-ops-notifications` channel is not being reviewed actively, and
presumably has not been for a long time. We should remove it.

The reasoning is relatively clear. When looking into the channel, we find a
large amount of almost completely unactionable problems with absolutely no
cross-referencing or other useful information. An issue was made to improve
this situation in Olli in August 2021, but has received no traction, see
python-discord/olli#12.

To further drive the point home, over the past few days, I have added a webhook
that periodically sends defamatory statements whilst impersonating Joe, and
nobody has complained yet. I therefore ask we remove the channel.

jb3:
This PR introduces Vector for log shipping amongst other QoL changes for our monitoring stack.

Vector is a very lightweight (~50mb) logging agent that can accept logs from a range of sources (similar to Alloy) and ship them to a range of sinks (in our case Loki).

We have switched Kubewatch from a custom deployment that sent data to Discord to a Helm deployment that will send structured JSON to Vector which will in turn push it into the Loki sink.

This allows us to keep track of events within our Kubernetes cluster over the long term in a more structured and useful fashion than just pumping information into a Discord channel.

Closes #347.


ChrisLovering (approved):
lgtm

PR #664, opened 2026-05-25
Add Alloy support to Rocky nodes

jb3:
- Add Rocky repositories for Grafana projects, so we can install Alloy
- Set new binary permissions for Alloy to allow reading of all files regardless of read permissions
- Enable live debugging features in Alloy
- Pull out systemd unit as a new label for better filtering in Loki

### DIFF
--- a/ansible/roles/alloy/tasks/main.yml
+++ b/ansible/roles/alloy/tasks/main.yml
@@ -1,14 +1,26 @@
 ---
 
-- name: Add Grafana apt repository with key
+- name: Add Grafana apt repository (Rocky)
+  yum_repository:
+    name: grafana
+    description: Grafana Repository
+    baseurl: "{{ alloy_rocky_grafana_repository }}"
+    gpgcheck: true
+    gpgkey: "{{ alloy_rocky_grafana_signing_key }}"
+  when: ansible_facts["distribution"] == "Rocky"
+  tags:
+    - role::alloy
+
+- name: Add Grafana apt repository with key (Debian)
   deb822_repository:
     name: grafana
     types: deb
-    uris: "{{ alloy_grafana_repository }}"
+    uris: "{{ alloy_debian_grafana_repository }}"
     state: present
     suites: [stable]
     components: [main]
-    signed_by: "{{ alloy_grafana_signing_key }}"
+    signed_by: "{{ alloy_debian_grafana_signing_key }}"
+  when: ansible_facts["distribution"] == "Debian"
   tags:
     - role::alloy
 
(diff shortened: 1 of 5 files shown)


ansible/roles/alloy/tasks/main.yml
```diff
@@ -1,14 +1,26 @@
...
 
-- name: Add Grafana apt repository with key
+- name: Add Grafana apt repository (Rocky)
+  yum_repository:
+    name: grafana
+    description: Grafana Repository
+    baseurl: "{{ alloy_rocky_grafana_repository }}"
+    gpgcheck: true
+    gpgkey: "{{ alloy_rocky_grafana_signing_key }}"
+  when: ansible_facts["distribution"] == "Rocky"
```
jchristgit, line 10: `ansible_facts["X"]` is equivalent to `ansible_X`, so `ansible_distribution` here

jchristgit: (see `ansible localhost -m setup`)

ansible/roles/alloy/tasks/main.yml
```diff
@@ -61,3 +73,39 @@
...
+  when: alloy_extra_groups | length > 0
+  tags:
+    - role::alloy
+  notify:
+    - Restart the alloy service
+
+# We need to add cap_dac_read_search=+ep to the Alloy binary.
+
+- name: Get Alloy binary path
+  command: "which alloy"
```
jchristgit, line 91: Do you know if ansible sanitizes `PATH` (or does not preserve it from the user it's escalating from)? If not, something malicious modifying `PATH` on the profile of whoever deploys could cause this to set `cap_dac_read_search` on the binary.

ansible/roles/alloy/tasks/main.yml
```diff
@@ -61,3 +73,39 @@
...
+
+- name: Get the current capabilities of the Alloy binary
+  command: "getcap {{ alloy_binary_path.stdout }}"
+  register: alloy_getcap_output
+  changed_when: false
+  tags:
+    - role::alloy
+
+- name: Set capabilities on the Alloy binary
+  command: "setcap cap_dac_read_search=ep {{ alloy_binary_path.stdout }}"
```
jchristgit, line 105: Can't this be done via systemd?

`man systemd.exec`, see `AmbientCapabilities`. That way we can remove this whole dance and just set it in the service, no?

ansible/roles/alloy/templates/config.alloy.j2
```diff
@@ -4,9 +4,35 @@ logging {
   level = "info"
 }
 
+livedebugging {
+  enabled = true
+}
+
```
jchristgit, line 10: Does this not expose sensitive information in some way that I'm missing? Judging by the Grafana docs, this relates to the Alloy UI. I haven't followed the deployment - do we expose that somewhere?

jchristgit (changes requested):

ansible/roles/alloy/tasks/main.yml
jb3: Confirmed privately this is because it's deprecated.

jchristgit: YAML continues its war on software. Sack of rice fell over. Chris Lovering hit by 5 gold bars stored in cupboard above bed. More at 12.

PR #666, opened 2026-05-27
Move mTLS CA cert mount out of /etc/nginx to avoid agent backup failure

ChrisLovering:


jchristgit (commented):
First commit LGTM. Second commit I have no idea. I think that's @jb3's area of influence.

jb3 (approved):

PR #671, opened 2026-07-02
Prevent flapping alerts

jchristgit:


shtlrs (approved):
I'm personally fine with this.

Idk how frequently Chris and Joe would want these to be, but we can sit in the "middle" and have it page every 10 mins

